PR #3, opened 2019-10-30
Access rights mgmt

tatut:
Added:
- user information to Datomic schema.
- new admin section
- role checks

### DIFF
--- a/app/backend/src/clj/teet/admin/admin_commands.clj
+++ b/app/backend/src/clj/teet/admin/admin_commands.clj
@@ -1,6 +1,7 @@
 (ns teet.admin.admin-commands
   (:require [teet.db-api.core :as db-api]
-            [datomic.client.api :as d]))
+            [datomic.client.api :as d]
+            [teet.user.user-roles :as user-roles]))
 
 (defmethod db-api/command! :admin/create-user [{conn :conn} {:user/keys [id person-id roles]}]
   (d/transact conn
@@ -8,3 +9,6 @@
                           :user/person-id person-id
                           :user/roles roles}]})
   :ok)
+
+(defmethod db-api/command-authorization :admin/create-user [{user :user} _]
+  (user-roles/require-role user :admin))
--- a/app/backend/src/clj/teet/admin/admin_commands.clj
+++ b/app/backend/src/clj/teet/admin/admin_commands.clj
@@ -1,0 +1,10 @@
+(ns teet.admin.admin-commands
+  (:require [teet.db-api.core :as db-api]
+            [datomic.client.api :as d]))
+
+(defmethod db-api/command! :admin/create-user [{conn :conn} {:user/keys [id person-id roles]}]
+  (d/transact conn
+              {:tx-data [{:user/id id
+                          :user/person-id person-id
+                          :user/roles roles}]})
+  :ok)
--- a/app/backend/src/clj/teet/admin/admin_commands.clj
+++ b/app/backend/src/clj/teet/admin/admin_commands.clj
@@ -1,14 +1,20 @@
 (ns teet.admin.admin-commands
   (:require [teet.db-api.core :as db-api]
             [datomic.client.api :as d]
-            [teet.user.user-roles :as user-roles]))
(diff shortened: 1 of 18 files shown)


app/backend/src/clj/teet/admin/admin_commands.clj
```diff
@@ -0,0 +1,20 @@
+(ns teet.admin.admin-commands
+  (:require [teet.db-api.core :as db-api]
+            [datomic.client.api :as d]
+            [teet.user.user-roles :as user-roles]
+            [taoensso.timbre :as log]))
+
+(defn- new-user []
+  {:user/id (java.util.UUID/randomUUID)
```
outamaa: That's some `teet.util/common/somesuch` material right there. Gotta have the `uuid` function. :)

app/backend/src/clj/teet/admin/admin_commands.clj
```diff
@@ -0,0 +1,20 @@
...
+   :user/roles [:user]})
+
+(defmethod db-api/command! :admin/create-user [{conn :conn} user-data]
+  (let [user (merge (new-user)
+                    (select-keys user-data [:user/id :user/person-id :user/roles]))]
+    (log/info "CREATE USER: " user)
```
outamaa: Here's a good `log/event` candidate in the future.

app/backend/src/clj/teet/admin/admin_queries.clj
```diff
@@ -0,0 +1,12 @@
+(ns teet.admin.admin-queries
+  (:require [teet.db-api.core :as db-api]
+            [teet.user.user-roles :as user-roles]))
+
+(defmethod db-api/query :admin/list-users [{db :db} _]
+  {:query '[:find (pull ?e [*])
+            :where [?e :user/id _]]
+   :args [db]
+   :result-fn (partial mapv first)})
```
outamaa: Is `(partial mapv first)` something that we want to do basically every time and then possibly something else as well?

app/backend/src/clj/teet/admin/admin_commands.clj
```diff
@@ -0,0 +1,20 @@
...
+    (d/transact conn
+                {:tx-data [user]})
+    :ok))
+
+(defmethod db-api/command-authorization :admin/create-user [{user :user} _]
```
outamaa: I'd rather see the authorization closely tied to the command/query definitions than as multimethods. I guess we can have a convention of having the authorization defined next to the command/query definition, but I don't immediately see what additional value using multimethods adds here.

app/backend/src/clj/teet/environment.clj
```diff
@@ -84,5 +84,5 @@
 
 (defn check-site-password [given-password]
   (let [actual-pw (config-value :auth :basic-auth-password)]
-    (and (some? actual-pw)
-         (= given-password actual-pw))))
+    (or (nil? actual-pw)
```
outamaa: 👍

app/common/src/cljc/teet/user/user_roles.cljc
```diff
@@ -0,0 +1,24 @@
...
+  ([{roles :user/roles :as user} role-or-roles]
+   (cond
+     (nil? role-or-roles)
+     true
+
+     (keyword? role-or-roles)
+     (contains? roles role-or-roles)
+
+     :else
+     (some #(has-role? user %) role-or-roles))))
```
outamaa: Ok, so here the sematics are "needs to have at least one of the listed roles". Maybe update the docstring "Returns true if user has the *single* given role *or one of the multiple given roles*"

app/backend/src/clj/teet/admin/admin_commands.clj
tatut: Well that is an issue for another PR, the query/command authorization split is existing.

app/backend/src/clj/teet/admin/admin_queries.clj
tatut: Not all queries will `pull` maps, but it is a common one for sure.

PR #4, opened 2019-10-30
TEET-226 Logging improvements

outamaa:
- Wrapping all logging related stuff (e.g. taoensso.timbre and datomic.ion.cast) in teet.log and re-exporting necessary stuff from there.
- Adding user id to all cast events, if available.
- Obtaining the user id from Timbre context.

The wrapping is to 1) enable having all possible logging with a single require and 2) discourage using infof, errorf etc which in my experience are just waiting to be called with a wrong number of arguments.


app/backend/src/clj/teet/log.clj
```diff
@@ -0,0 +1,84 @@
+(ns teet.log
```
tatut: We use timbre logging on the frontend as well. Should we unify and have the wrapper there as well?

outamaa: Fine by me. At first we could just expose the blessed set of functions from timbre. Later we could maybe have a frontend logging endpoint where we'd send `log/error`s to backend for being further logged into CloudWatch.

PR #8, opened 2019-11-15
THK import Datomic

tatut:
Project data is now in Datomic.

PostGIS still contains all the geometry data and an "entity" table that is generic.


app/backend/resources/schema.edn
```diff
@@ -304,14 +241,113 @@
...
            :db/cardinality :db.cardinality/one
-           :db/doc "THK project id"
-           :db/unique :db.unique/identity}]]}]
+           :db/doc "Activity status enum ref"}]]}]
+
+
 ;; Project
-;; - Lifecycle phase
-;;   - Activity (formerly known as phase)
+;; - Lifecycle activity
```
outamaa: Lifecycle phase. Or maybe these should be removed now.

PR #9, opened 2019-11-19
TEET-246 Implement feature flags

erno:
https://jira.mkm.ee/browse/TEET-246

Feature flag config changes made also to mnt-teet-private repo


app/backend/src/clj/teet/environment.clj
```diff
@@ -20,6 +21,20 @@
 
 (def config (atom init-config))
 
+;; "road-information-view;component-view"
+(defn parse-enabled-features [ssm-param]
+  (->> (str/split ssm-param #",")
+       (remove empty?)
```
tatut: could use `str/blank?` as remove predicate

PR #15, opened 2020-02-17
Teet 339 select related entities

jkorpelin:
Added functionality to edit related features.

Some parts will still be quite messy since all the showing of lists were heavily bundled with the setup wizard

### DIFF
--- a/app/backend/src/clj/teet/project/project_commands.clj
+++ b/app/backend/src/clj/teet/project/project_commands.clj
@@ -133,7 +133,7 @@ and cadastral units"
 
 (defcommand :thk.project/update-restrictions
   {:doc "Update project related restrictions"
-   :context {:keys [user db conn]}
+   :context {:keys [user db]}
    :payload {:keys [restrictions project-id]}
    :project-id [:thk.project/id project-id]
    :authorization {:project/project-info {:eid [:thk.project/id project-id]
@@ -142,7 +142,7 @@ and cadastral units"
 
 (defcommand :thk.project/update-cadastral-units
   {:doc "Update project related cadastral-units"
-   :context {:keys [user db conn]}
+   :context {:keys [user db]}
    :payload {:keys [cadastral-units project-id]}
    :project-id [:thk.project/id project-id]
    :authorization {:project/project-info {:eid [:thk.project/id project-id]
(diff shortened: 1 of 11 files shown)


app/backend/src/clj/teet/project/project_commands.clj
```diff
@@ -95,46 +97,83 @@ and cadastral units"
...
+            (for [id-to-remove to-be-removed]
+              [:db/retract project-eid
+               link-attribute-kw id-to-remove])
+            (for [id-to-add to-be-added]
+              [:db/add project-eid
+               link-attribute-kw id-to-add])))))
+
+(defcommand :thk.project/update-restrictions
+  {:doc "Update project related restrictions"
+   :context {:keys [user db conn]}
```
tatut: dont need to destructure `conn` from context anymore, same in the below command

app/frontend/src/cljs/teet/project/project_controller.cljs
```diff
@@ -220,17 +220,44 @@
       (->NavigateToStep step-label))))
 
 (extend-protocol t/Event
-  FetchRelatedFeaturesResponse
-  (process-event [{:keys [result-path geojson-path response]} app]
-    (let [geojson (js/JSON.parse response)
+  FetchFeatureCandidatesResponse
+  (process-event [{:keys [candidate-type response]} app]
+    (let [result-path (if (= candidate-type :restrictions)
```
tatut: these could be put to a "config map" based on candidate type and destructured from there

`(def candidate-paths {:restrictions {:result-path ...} :cadastral-units {:result-path ...}})` 

and uses `(let [{:keys [result-path ...]} (candidate-paths candidate-type)`

app/frontend/src/cljs/teet/project/project_layers.cljs
```diff
@@ -173,20 +187,26 @@
...
 
-(defn selected-restrictions [{{checked-restrictions-geojson :checked-restrictions-geojson
+(defn selected-restrictions [{{:keys [query]} :app
+                              {checked-restrictions-geojson :checked-restrictions-geojson
                                setup-step :setup-step} :project}]
   (when (and checked-restrictions-geojson
-             (= setup-step "restrictions"))
+             (or
+               (= (:configure query) "restrictions")
+               (= setup-step "restrictions")))
```
outamaa: This pattern is repeated quite a few times. I guess this has to do with getting things working in both the wizard and the new views?

jkorpelin: Yes, would be great if this could somehow be made more clear.

PR #25, opened 2020-03-17
Frontend version desync / deployment break detection: detect any change to versions, don't try to track git revisions

erno:
Based on Ville's idea from yesterday

### DIFF
--- a/app/frontend/src/cljs/teet/common/common_controller.cljs
+++ b/app/frontend/src/cljs/teet/common/common_controller.cljs
@@ -389,7 +389,14 @@
                  (reset! version-seen-on-startup commit))
                
                (js/setTimeout #(poll-version e!) poll-timeout-ms)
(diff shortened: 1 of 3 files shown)


app/frontend/src/cljs/teet/common/common_controller.cljs
```diff
@@ -373,25 +372,31 @@
...
 (defn poll-version
   "Polls whether the client's version of the app is the same as the one
   currently deployed. If not, shows a warning that a new version is
   being, or has been, deployed."
   [e!]
   (-> (fetch-deploy-status)
       (.then (fn [{:strs [commit status]}]
+               (when (and (some? commit)
+                          (nil? @version-seen-on-startup))
+                 (reset! version-seen-on-startup commit))
```
outamaa: Nitpicking, but you could just `(swap! version-seen-on-startup (fn [version] (or version commit)))` to avoid the extra deref.

PR #30, opened 2020-04-07
Add minimal setup for db tests

outamaa:
This PR adds an MVP version of DB tests.

### DIFF
--- a/app/backend/src/clj/user.clj
+++ b/app/backend/src/clj/user.clj
@@ -4,6 +4,7 @@
             [teet.log :as log]
             [teet.main :as main]
             [teet.environment :as environment]
+            [teet.test.utils :as tu]
             [teet.thk.thk-integration-ion :as thk-integration]
             [teet.user.user-db :as user-db]
             [clojure.string :as str])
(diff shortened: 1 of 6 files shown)


app/backend/src/clj/user.clj
```diff
@@ -4,6 +4,7 @@
             [teet.log :as log]
             [teet.main :as main]
             [teet.environment :as environment]
+            [teet.test.utils :as tu]
             [teet.thk.thk-integration-ion :as thk-integration]
             [teet.user.user-db :as user-db]
```
tatut, line 9: fix lint warnings

app/backend/src/clj/user.clj
```diff
@@ -230,37 +231,15 @@
...
-  ([command args]
-   (if-let [user-id @logged-in-user-id]
-     (local-command user-id command args)
-     (log/error "Not logged in! Call user/local-login with an existing user id to log in.")))
-
-  ([user-id command args]
-   (db-api-handlers/raw-command-handler (action-ctx user-id)
-                                        {:payload args
-                                         :command command})))
+(intern 'user
```
tatut, line 234: why intern instead of `def`?

app/backend/test/teet/test/utils.clj
```diff
@@ -0,0 +1,83 @@
...
+  (log/info "Reloading local config.")
+  (environment/load-local-config!))
+
+(defn with-db [f]
+  (log/redirect-ion-casts! :stderr)
+  (let [test-db-name (str "test-db-" (System/currentTimeMillis))]
+    (log/info "Creating database " test-db-name)
+    (swap! environment/config assoc-in [:datomic :db-name] test-db-name)
+    (d/create-database (d/client (environment/config-value :datomic :client)) {:db-name test-db-name})
+    (f)
```
tatut, line 25: I think this should be wrapped in exception handling to avoid leaving databases.

```clojure
(try 
  (f)
  (finally
    (d/delete...))
```

app/backend/src/clj/user.clj
outamaa: Yeah true, since they're not macros, might as well just use a def.

app/backend/test/teet/test/utils.clj
outamaa: I tested this, and the test framework seems to catch the exceptions and allow the fixtures to finish without a `try` block:
```
INFO 2020-04-07 10:09:46.777 [] teet.test.utils:26 - Deleting database  test-db-1586243375591
INFO 2020-04-07 10:09:47.188 [] teet.test.utils:16 - Reloading local config.
[...]
ERROR in teet.system-test/throw-stuff (system_test.clj:14)
Uncaught exception, not in assertion.
Exception: clojure.lang.ExceptionInfo: thrown
[...]
```

PR #32, opened 2020-04-09
Teet 428 landacquisition

jkorpelin:


app/backend/src/clj/teet/land/land_commands.clj
```diff
@@ -0,0 +1,64 @@
+(ns teet.land.land-commands
+  (:require [teet.db-api.core :as db-api :refer [defcommand]]
+            [datomic.client.api :as d]
+            [teet.meta.meta-model :refer [modification-meta creation-meta] :as meta-model]
```
tatut, line 4: fix lint warnings

PR #39, opened 2020-05-05
TEET 531, improve part 3 / part 4 parsing from property register responses

erno:
Now provides much more complete information.

### DIFF
--- a/app/backend/dev-src/user.clj
+++ b/app/backend/dev-src/user.clj
@@ -236,7 +236,7 @@
   ;; when testing submit / approve workflow this lets you walk back the status
   (let [status-res (d/pull (db) '[:activity/status] act-id)
         curr-status (-> status-res :activity/status :db/ident)]
-    (when (curr-status)
+    (when curr-status
       (tx
        {:db/id act-id
         :activity/status :activity.status/in-progress}))))
--- a/app/backend/dev-src/user.clj
+++ b/app/backend/dev-src/user.clj
@@ -236,9 +236,10 @@
   ;; when testing submit / approve workflow this lets you walk back the status
   (let [status-res (d/pull (db) '[:activity/status] act-id)
         curr-status (-> status-res :activity/status :db/ident)]
-    (tx
-     {:db/id act-id
-      :activity/status :activity.status/in-progress})))
+    (when (curr-status)
+      (tx
+       {:db/id act-id
+        :activity/status :activity.status/in-progress}))))
 
 
 ;;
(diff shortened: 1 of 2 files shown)


app/backend/dev-src/user.clj
```diff
@@ -236,9 +236,10 @@
   ;; when testing submit / approve workflow this lets you walk back the status
   (let [status-res (d/pull (db) '[:activity/status] act-id)
         curr-status (-> status-res :activity/status :db/ident)]
-    (tx
-     {:db/id act-id
-      :activity/status :activity.status/in-progress})))
+    (when (curr-status)
```
tatut, line 239: I think calling `curr-status` as a function will fail (it's either a kw or nil)

app/backend/src/clj/teet/integration/x_road.clj
```diff
@@ -155,16 +155,73 @@
               :omandiosa_suurus (omandi-get ox :a:omandiosa_suurus)})
            owner-xml-seq)}))
 
+(defn d-jagu34* [key children sequential-vals?]
```
tatut, line 158: this fn body could use a docstring and some comments... pretty terse and not obvious what it does

app/backend/src/clj/teet/integration/x_road.clj
```diff
@@ -155,16 +155,73 @@
...
+                      leaves))
+        maybe-vectorise (if sequential-vals?
+                          (fn [val] [val])
+                          identity)]
+    
+    {key (maybe-vectorise
+          (merge recursed-map
+                 leaves-map))}))
 
 (defn d-jagu34 [k-xml p1 p2]
```
tatut, line 186: fix lint warning

app/backend/src/clj/teet/integration/x_road.clj
```diff
@@ -155,16 +155,73 @@
...
+
+(defn active-jagu34-only [k-resp]
+  ;; prunes "lopetatud" records from the maps in the seq under :jagu3 and :jagu4.
+  ;; {:jagu3 [... {:oiguse_seisund "L", ..} ...]
+  (let [filter-fn #(filter (comp (partial not= "L") :oiguse_seisund) %)]
+    (-> k-resp
+        (update :jagu3 filter-fn)
+        (update :jagu4 filter-fn))))
+
+(defn parse-kande-tekst [kande-tekst-str]
```
tatut, line 217: If we only need to show the HTML content, I would try just not parsing it and showing it as is with `dangerouslySetInnerHTML` I don't think any scripts are coming from property register.

PR #43, opened 2020-05-29
TEET-605: Allow activity delete if it doesn't have a procurement number

outamaa:
Besides what it says above, the PR also includes
- Migration of procurement nr and id into their own attributes from integration info.
- Relevant changes to THK/TEET integration code.


app/backend/src/clj/teet/migration/activity_procurement_info.clj
```diff
@@ -0,0 +1,21 @@
...
+     conn
+     {:tx-data (for [[activity-id integration-info] activities
+                     :let [info (parse integration-info)
+                           procurement-id (:activity/procurementid info)
+                           procurement-nr (:activity/procurementno info)]
+                     :when procurement-id]
+                 (merge {:db/id activity-id
+                         :activity/procurement-id procurement-id}
+                        (when procurement-nr
+                          {:activity/procurement-nr procurement-nr})))})))
```
outamaa, line 21: I'll push a change that will simplify this a bit once I've ensured all tests pass.

PR #45, opened 2020-06-07
TEET-576 Update data model, add manager history to activities when fetching project

outamaa:
https://jira.mkm.ee/browse/TEET-576

- Added `:activity/manager` attribute
- Added project-manager-histories to activities when fetching project
- Added user/add-activity-manager for dev convenience


app/backend/src/clj/teet/activity/activity_db.clj
```diff
@@ -21,3 +22,100 @@
...
+
+
+(defn manager-histories-by-activity [manager-transactions users-by-id]
+  (->> manager-transactions
+       (group-by :activity)
+       (cu/map-vals (comp (partial map #(update % :manager users-by-id))
+                          manager-transactions->manager-periods
+                          (partial sort-by :modified-at)))))
+
+(defn get-activity-ids-of-project
```
tatut, line 68: is get verb necessary, this looks like a pure function.

It could be called `project-activity-ids` to reflect the return value instead.

app/backend/src/clj/teet/activity/activity_db.clj
```diff
@@ -21,3 +22,100 @@
               (and (not (.before date (:activity/estimated-start-date activity-dates)))
                    (not (.after date (:activity/estimated-end-date activity-dates)))))
             dates)))
+
```
erno, line 25: Cool to see d/history in action :)

outamaa: History in the making :badumtss:

app/backend/src/clj/teet/activity/activity_db.clj
outamaa: It queries the database? Which is unnecessary btw, since the ids could also be gathered from the project passed to update-activity-histories.

PR #46, opened 2020-06-16
Keep track of deleted status of geo features

erno:
This is done by checking which feature IDs have been previously
imported but are missing in newly imported SHP files.


tatut:
Should we update all the queries that return features to not include deleted ones? Except when explicitly fetched by id.

PR #47, opened 2020-06-22
Add audit logging support

tatut:


outamaa:
Looking good so far.

tatut:
Using AWS structured cloudwatch audit logging instead, they can be redirected to whatever persistent storage is needed.

PR #48, opened 2020-06-23
Add :file/upload-complete? and only return completed files

tatut:
- Add `:file/upload-complete?` attribute
- Mark file upload as complete after the S3 PUT request is done
- Only return upload complete files in file listing query

### DIFF
--- a/app/backend/src/clj/teet/file/file_commands.clj
+++ b/app/backend/src/clj/teet/file/file_commands.clj
@@ -71,6 +71,18 @@
                 ;; Take part after last underscore to skip over invalid metadata
                 (last (str/split n #"_")))))))
 
+(defcommand :file/upload-complete
+  {:doc "Mark file upload as complete"
+   :context {:keys [conn user db]}
+   :payload {id :db/id}
+   :pre [(file-db/own-file? db user id)]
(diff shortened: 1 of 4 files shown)


app/backend/src/clj/teet/file/file_commands.clj
```diff
@@ -71,6 +71,18 @@
...
+  {:doc "Mark file upload as complete"
+   :context {:keys [conn user db]}
+   :payload {id :db/id}
+   :pre [(file-db/own-file? db user id)]
+   ;; No need to check extra authorization again, as we check pre condition that this
+   ;; is the user's own uploaded file
+   :project-id nil
+   :authorization {}
+   :transact [{:db/id id
+               :file/upload-complete? true}]})
```
outamaa, line 84: So what happens if the upload actually completes but the client doesn't call this?

outamaa: Due to e.g. page reload.

tatut: the file is forever incomplete and will not show on the listing

tatut: but that can be queried for by searching for files that are created long ago but don't have upload-complete yet

PR #50, opened 2020-06-30
Remove Amazonica and direct AWS SDK dependencies

tatut:
Use Cognitect AWS API instead of including AWS Java SDK and Amazonica.

- Implemented all needed S3 operations in `teet.integration.integration-s3`

Tested file upload/download works as expected. Tested other operations from REPL.


tatut:
idk if this is worth it merging... the deps are reduced, but jackson databind still comes thru datomic ion dep

on the plus side, I like the cognitect AWS API and it is very nice to use (and has great introspection capabilities) so I would like to use that instead of amazonica going forward

app/backend/src/clj/teet/integration/integration_s3.clj
```diff
@@ -56,6 +89,138 @@
...
+            (.setTimeZone (java.util.TimeZone/getTimeZone "UTC")))]
+  (defn- timestamp [d]
+    (.format fmt d)))
+
+(defn- hex [bytes]
+  (str/join ""
+            (map #(format "%02x" %) bytes)))
+
+(defn- ->b [string]
+  (.getBytes string "UTF-8"#_"US-ASCII"))
```
outamaa, line 163: Nitpick: #_"US-ASCII"

outamaa: Is this used anywhere?

PR #54, opened 2020-08-07
Remove :user/roles attribute as deprecated

outamaa:
The PR removes :user/roles and associated code as deprecated. Pay special attention to TARA related code when reviewing.


app/backend/src/clj/teet/db_api/core.clj
```diff
@@ -133,8 +132,6 @@
                           :query args) -payload) ~-payload
                    ~-db (d/db (:conn ~-ctx))
                    ~-user (:user ~-ctx)
-                   ~-perms (when (:user ~-ctx)
-                             (permission-db/user-permissions ~-db [:user/id (:user/id (:user ~-ctx))]))
```
tatut, line 137: also remove require from ns form as it doesn't seem to be in use anymore

PR #58, opened 2020-09-02
TEET-817

tatut:
Fix add files to task modal.

### DIFF
--- a/app/frontend/src/cljs/teet/task/task_view.cljs
+++ b/app/frontend/src/cljs/teet/task/task_view.cljs
@@ -160,7 +160,24 @@
                                    :style {:margin-left "1rem"}}
            (tr [:buttons :confirm])]]]])]))
 
-(declare add-files-form)
+(defn- add-files-form [e! upload-progress close!]
+  (r/with-let [form (r/atom {})]
+    [:<>
+     [form/form {:e!              e!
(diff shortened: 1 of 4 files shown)


app/frontend/src/cljs/teet/ui/panels.cljs
```diff
@@ -198,10 +198,10 @@
 (defn button-with-modal [{:keys [modal-title button-component modal-component]}]
   (r/with-let [open? (r/atom false)
                open! #(reset! open? true)
-               close! #(reset! open? false)]
+               close! #(do (reset! open? false) nil)]
```
jkorpelin, line 201: The do seems unnecessary

tatut: It is necessary, I want to explicitly return nil

app/frontend/src/cljs/teet/task/task_view.cljs
```diff
@@ -160,16 +160,20 @@
                                    :style {:margin-left "1rem"}}
            (tr [:buttons :confirm])]]]])]))
 
+(declare add-files-form)
```
jkorpelin, line 163: Should the add-files-form just be moved up instead of declared?

PR #63, opened 2020-09-24
TEET-878

erno:


app/backend/src/clj/teet/thk/thk_export.clj
```diff
@@ -81,9 +81,9 @@
 
                  ;; TEET id for phase and activity
                  "phase_teetid"
-                 (str (:db/id lifecycle))
+                 (str (thk-mapping/uuid->number (:integration/id lifecycle)))
```
erno, line 84: Are these nil safe after the change / do they need to be ?

tatut: It is required that all things sent in this integration have that id.

It could be better checked and reported if not, but we want the export to fail if they aren't present.

tatut:
things to check
- time differs when running in cloud vs locally (locale setting)
- don't create new UUID for phases we haven't sent back (phases without activities)

PR #66, opened 2020-10-06
TEET-878 smaller UUID numbers

tatut:
Use only 64 bits in the UUID numbers as THK has column type int8.

Ids are created by secure random and checked against the current database that they are not in use.

### DIFF
--- a/app/backend/src/clj/teet/thk/thk_mapping.clj
+++ b/app/backend/src/clj/teet/thk/thk_mapping.clj
@@ -1,7 +1,8 @@
 (ns teet.thk.thk-mapping
   "Mapping of information between THK CSV and TEET attribute keywords"
   (:require [clojure.string :as str]
-            [clojure.java.io :as io])
+            [clojure.java.io :as io]
+            [datomic.client.api :as d])
   (:import (java.text SimpleDateFormat)))
 
 (defn- blank? [s]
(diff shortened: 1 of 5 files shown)


app/backend/src/clj/teet/thk/thk_mapping.clj
```diff
@@ -24,6 +25,24 @@
...
+
+(defn- used-integration-id? [db id]
+  (boolean
+   (seq
+    (d/q '[:find ?e :where [?e :integration/id ?v] :in $ ?v] db id))))
+
+(defn unused-random-small-uuid [db]
+  (first
+   (drop-while (partial used-integration-id? db)
+               (repeatedly random-small-uuid))))
```
outamaa, line 44: Entäs sitten kun meiltä alkaa loppua vapaat 64-bittiset numerot kesken? :ba-dum-tss:

outamaa (approved):
One can construct some hypothetical situations where the implementation could go awry, but given the amount of 64 bit numbers, and the ratio of used / unused ones we'd ever hope to achieve during the lifetime of the application, this is good enough.

app/backend/src/clj/teet/thk/thk_mapping.clj
```diff
@@ -24,6 +25,24 @@
           msb (.longValue (.shiftRight n 64))]
       (java.util.UUID. msb lsb))))
 
+(defonce random (java.security.SecureRandom.))
+
+(defn random-small-uuid
+  "Create a UUID that only has 64bits number. This is because THK
+  teet id numbers are only int8 sized and cannot fit a full UUID"
+  []
+  (number->uuid (.nextLong random)))
```
erno, line 34: A problem with the default Java SecureRandom used to be that it could hang for long times because it had a bad default randomness source on Linux (worked around by starting JVM with  -Djava.security.egd=file:/dev/urandom ). Hopefully this is fixed the JVM version used by Datomic...

erno (approved):

PR #68, opened 2020-10-08
TEET-658  - Meeting link addition to ics & notification implementation for past meeting editing

erno:


app/backend/src/clj/teet/meeting/meeting_commands.clj
```diff
@@ -204,15 +205,52 @@
...
+  (tx-ret
+   (vec
+    (for [to-user participants]
+      (notification-db/notification-tx
+       db
+       (do
+         (log/debug "to-user is" to-user)
+           
+         {:from from-user
+          :to to-user
```
erno, line 220: what if there are participants who are not registered in teet, will to list have those?

app/backend/src/clj/teet/meeting/meeting_commands.clj
```diff
@@ -204,15 +205,52 @@
...
+                          (environment/config-value :base-url)
+                          meeting project-eid)
+            ics (meeting-ics/meeting-ics {:meeting meeting
+                                          :meeting-link meeting-link
+                                          :cancel? false})
+            email-response
+            (integration-email/send-email!
+              {:from (environment/ssm-param :email :from)
+               :to to
+               :subject (str "TEET: koosoleku kutse " (meeting-model/meeting-title meeting) " / "
```
erno, line 236: this has to change for the past-meeting case because it's no longer an invitation

app/backend/src/clj/teet/meeting/meeting_commands.clj
```diff
@@ -221,34 +259,19 @@
...
-                                                                          :user/family-name]}]}] id)]
+                                                                          :user/family-name]}]}] meeting-eid)]
+    (assert (:db/id user))
     (cond
-      (date-in-past? (:meeting/end meeting))
-      {:error :no-emails-after-meeting-end}
+      (date-in-past? (:meeting/start meeting))
+      (historical-meeting-notify db meeting user
+                                 (meeting-db/participants db meeting-eid)
+                                 project-eid)
```
erno, line 268: add email sending (past meeting case) here

app/backend/src/clj/teet/meeting/meeting_commands.clj
erno: yes, but it works out since those users also seem to have user objects in db

PR #72, opened 2020-10-16
TEET-722 Add audit? option to actions, audit log privilege granting commands

outamaa:
This PR adds the option to add audit logging to actions (that is, commands and queries) by adding `audit? true` to the action options. This results in the attempt to call the given action being logged into audit log.


erno (approved):
LGTM

PR #74, opened 2020-10-20
Add more THK export logging

erno:


app/backend/src/clj/teet/thk/thk_integration_ion.clj
```diff
@@ -171,7 +174,9 @@
            export-projects
            csv->file
            write-file-to-s3)
-    "{\"success\": true}"
+    (do
```
tatut, line 177: remove redundant do

PR #79, opened 2020-11-02
TEET-907: move allowed file suffixes to param store

erno:


app/backend/test/teet/comment_commands_test.clj
```diff
@@ -235,7 +236,9 @@
       "task is assigned to carla")
   (tu/local-login tu/mock-user-carla-consultant)
   (with-redefs [;; Mock out URL generation (we don't use it for anything)
```
erno, line 238: note to self: move the default suffixes to test fixture

app/backend/test/teet/file/file_commands_test.clj
```diff
@@ -3,6 +3,7 @@
             [teet.file.file-model :as file-model]))
 
 (deftest validate-document
+  (reset! file-model/upload-allowed-file-suffixes #{"doc" "xslx" "mp4" "png"})
```
erno, line 6: note to self: move the default suffixes to test fixture

app/backend/src/clj/teet/environment.clj
```diff
@@ -118,13 +119,34 @@
...
 (defn- load-ssm-config! [base-config]
   (let [old-config @config
         new-config (merge base-config
                           (resolve-ssm-parameters teet-ssm-config))]
     (when (not= old-config new-config)
       (log/info "Configuration changed")
-      (reset! config new-config))))
+      (reset! config new-config)
+      ;; do after config reset, don't block other config reload in case of errors
+      (file-model/apply-attachment-policy! (map :Name (describe-all-the-params!))))))
```
tatut, line 149: I don't understand the rationale for `describe-all-the-params!` why isn't the allowed file suffix list a regular parameter like others?

tatut: seems more difficult to configure instead of just having a comma separated list of suffixes that are allowed

tatut:
Move to single SSM parameter, like other configs, so it's easier to manage. Will merge the tests from here.

tatut:
Closing, but keep branch around if we need to come back to more configurable file validation policy.

PR #81, opened 2020-11-05
TEET-822

outamaa:
https://jira.mkm.ee/browse/TEET-822


jkorpelin (approved):
Looks good to me. Like you noticed in the demo, the disabled-buttons? can be cleaned out from the project navigator.

PR #85, opened 2020-11-11
TEET-905 use :integration/id for project geometries

tatut:
Create :integration/id for THK projects that can be used as stable id that doesn't need special handling when doing backup/restore.


jkorpelin (approved):
LGTM

PR #89, opened 2020-11-16
TEET-944 - show joint ownership

erno:


erno:
This is missing the visual refinements to owner-info modal discussed  on the tickets, those can be done as a separate PR.

outamaa (approved):

PR #92, opened 2020-11-19
Teet 924 notification time

jkorpelin:
Adds notification info to frontend about notification times


outamaa (approved):
LGTM after quick browsing.

PR #102, opened 2020-12-01
TEET-615 adds error handling for creating applications outside activi…

jkorpelin:
…tes dates, also adds some command tests for validation applications

### DIFF
--- a/app/backend/test/cooperation/cooperation_commands_test.clj
+++ b/app/backend/test/cooperation/cooperation_commands_test.clj
@@ -1,0 +1,62 @@
+(ns ^:db cooperation.cooperation-commands-test
+  (:require [clojure.test :refer [deftest is testing] :as t]
+            [teet.test.utils :as tu]
+            [teet.util.date :as dateu]
+            teet.task.task-commands
+            [teet.util.datomic :as du]))
+
+(t/use-fixtures
+  :each
+  tu/with-environment
+  (tu/with-db)
+  tu/with-global-data)
+
+(deftest test-for-third-party
+  (tu/local-login tu/mock-user-boss)
+  (let [project-id (:thk.project/id (du/entity (tu/db) (tu/->db-id "p1")))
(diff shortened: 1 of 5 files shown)


app/backend/test/cooperation/cooperation_commands_test.clj
```diff
@@ -0,0 +1,62 @@
...
+                     :cooperation.application/date valid-date-for-application}
+        invalid-date-application (assoc application :cooperation.application/date invalid-date-for-application)
+        application-payload {:thk.project/id project-id
+                             :cooperation.3rd-party/name third-party-name
+                             :application application}
+        invalid-third-party-application-payload {:thk.project/id project-id
+                                                 :cooperation.3rd-party/name "No such third party exists"
+                                                 :application application}
+        res (tu/local-command :cooperation/create-application
+                              application-payload)]
```
outamaa, line 42: The `global-data` stuff can be used if one wants to have the API calls within the `testing` body without nesting lots of `let`s.

app/common/src/cljc/teet/util/date.cljc
```diff
@@ -25,13 +22,25 @@
...
+         .getTime
+         (+ (* days 24 60 60 1000))
+         Date.)))
+
+#?(:clj
+   (defn dec-days [date days]
+     (-> date
+         .getTime
+         (- (* days 24 60 60 1000))
+         Date.)))
```
outamaa, line 43: Since we're using `clj-time`, here's another possible implementation:

```
(defn inc-days [date days]
  (-> date
      tc/from-date
      (t/plus (t/days days))
      tc/to-date))
```
where `t = clj-time.core` and `tc = clj-time.coerce`. More of an FYI in case we need to do some heavier date juggling in the future.

outamaa (approved):
Approved, but `application-matched-activity-id` should have a better docstring (I assume it's unfinished) and maybe a more descriptive name.

PR #103, opened 2020-12-01
Teet 926 unit tests

aleksievelizhanin:
1. add unit tests for Meeting pdf
2. fix Meeting pdf file name to the one containing the titlte

### DIFF
--- a/app/backend/src/clj/teet/meeting/meeting_queries.clj
+++ b/app/backend/src/clj/teet/meeting/meeting_queries.clj
@@ -331,16 +331,17 @@
    :context {:keys [db user]}
    :args {id :db/id meeting :meeting}
    :project-id (project-db/meeting-project-id db id)
-   :authorization {:meeting/download-attachment {:db/id id :link  :meeting/organizer-or-reviewer}}}
-      ^{:format :raw}
-          {:status  200
+   :authorization {:meeting/download-attachment {:db/id id
+                                                 :link :meeting/organizer-or-reviewer}}}
+          ^{:format :raw}
+          {:status 200
            :headers {"Content-Disposition" (str "inline; filename=meeting_" (meeting-model/meeting-title meeting) ".pdf")
-                     "Content-Type"        "application/pdf"}
-           :body    (ring-io/piped-input-stream
-                      (fn [ostream]
-                        (try
-                          (pdf-export/hiccup->pdf
-                            (meeting-pdf/meeting-pdf db user id)
-                            ostream)
-                          (catch Exception e
-                            (log/error e "Exception while generating meeting PDF")))))})
+                     "Content-Type" "application/pdf"}
+           :body (ring-io/piped-input-stream
+                   (fn [ostream]
+                     (try
+                       (pdf-export/hiccup->pdf
+                         (meeting-pdf/meeting-pdf db user id)
+                         ostream)
+                       (catch Exception e
+                         (log/error e "Exception while generating meeting PDF")))))})
--- a/app/backend/src/clj/teet/meeting/meeting_queries.clj
+++ b/app/backend/src/clj/teet/meeting/meeting_queries.clj
@@ -333,15 +333,15 @@
    :project-id (project-db/meeting-project-id db id)
    :authorization {:meeting/download-attachment {:db/id id
                                                  :link :meeting/organizer-or-reviewer}}}
-          ^{:format :raw}
-          {:status 200
-           :headers {"Content-Disposition" (str "inline; filename=meeting_" (meeting-model/meeting-title meeting) ".pdf")
-                     "Content-Type" "application/pdf"}
-           :body (ring-io/piped-input-stream
-                   (fn [ostream]
-                     (try
-                       (pdf-export/hiccup->pdf
-                         (meeting-pdf/meeting-pdf db user id)
-                         ostream)
-                       (catch Exception e
-                         (log/error e "Exception while generating meeting PDF")))))})
+  ^{:format :raw}
+  {:status 200
+   :headers {"Content-Disposition" (str "inline; filename=meeting_" (meeting-model/meeting-title meeting) ".pdf")
+             "Content-Type" "application/pdf"}
+   :body (ring-io/piped-input-stream
+          (fn [ostream]
+            (try
+              (pdf-export/hiccup->pdf
+               (meeting-pdf/meeting-pdf db user id)
+               ostream)
+              (catch Exception e
+                (log/error e "Exception while generating meeting PDF")))))})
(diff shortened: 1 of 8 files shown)


app/backend/src/clj/teet/meeting/meeting_queries.clj
```diff
@@ -332,15 +332,15 @@
...
-             "Content-Type" "application/pdf"}
-   :body (ring-io/piped-input-stream
-          (fn [ostream]
-            (try
-              (pdf-export/hiccup->pdf
-               (meeting-pdf/meeting-pdf db user id)
-               ostream)
-              (catch Exception e
-                (log/error e "Exception while generating meeting PDF")))))})
+            ^{:format :raw}
```
tatut, line 335: remove extra indentation in this form

app/backend/src/clj/teet/project/project_db.clj
```diff
@@ -97,6 +97,9 @@
           [:activity/_meetings 0
            :db/id]))
 
+(defn meeting-title [db meeting-id]
+  (:meeting/title (du/entity db meeting-id)))
```
tatut, line 101: this should include the meeting number as well, as many meetings can have the same title

app/backend/test/teet/meeting/meeting_pdf_test.clj
```diff
@@ -0,0 +1,26 @@
...
+     :meeting/end (java.util.Date. (+ now (* 1000 60 60)))
+     :meeting/organizer tu/mock-user-manager}))
+
+(deftest print-meeting-pdf
+  (let [new-meeting-id (meeting-commands-test/create-meeting! (tu/->db-id "p1-lc1-act1") (test-meeting))]
+  (tu/local-login tu/mock-user-boss)
+  (testing
+    "PDF has correct title"
+    (let [pdf (teet.meeting.meeting-pdf/meeting-pdf (tu/db) (tu/logged-user) new-meeting-id)]
+      (is (= (hiccup.core/html (nth (last (last pdf)) 2)) title-fo-block))))))
```
tatut, line 26: this access path looks a little brittle and the test is very specific and likely to break if the PDF is changed at all.

there's no easy way to test the PDF structure without being either too specific or very loose (just checking some text is present), so maybe that's fine

app/backend/test/teet/meeting/meeting_pdf_test.clj
```diff
@@ -0,0 +1,26 @@
...
+(defn test-meeting []
+  (let [now (System/currentTimeMillis)]
+    {:meeting/title "Test meeting for PDF report"
+     :meeting/location "Somewhere inside the test runner"
+     :meeting/start (java.util.Date. (+ now (* 1000 60)))
+     :meeting/end (java.util.Date. (+ now (* 1000 60 60)))
+     :meeting/organizer tu/mock-user-manager}))
+
+(deftest print-meeting-pdf
+  (let [new-meeting-id (meeting-commands-test/create-meeting! (tu/->db-id "p1-lc1-act1") (test-meeting))]
```
tatut, line 21: you need to call local-login before creating the meeting

app/backend/test/teet/meeting/meeting_pdf_test.clj
tatut, line 26: require `hiccup.core` namespace

app/backend/test/teet/meeting/meeting_pdf_test.clj
```diff
@@ -0,0 +1,26 @@
+(ns ^:db teet.meeting.meeting-pdf-test
+  (:require teet.meeting.meeting-pdf
+            [teet.meeting.meeting-commands-test :as meeting-commands-test]
+            [teet.test.utils :as tu]
+            [clojure.test :refer [deftest is testing use-fixtures]]
+            [datomic.client.api :as d]))
```
tatut, line 6: remove unused `datomic.client.api` require

app/backend/test/teet/meeting/meeting_pdf_test.clj
aleksievelizhanin: At least there is a test that verify current PDF structure (title). If PDF changes the test will notify if titlte is altered. So, I guess it is rather good than bad to have this path.

app/backend/src/clj/teet/meeting/meeting_queries.clj
```diff
@@ -332,15 +332,15 @@
...
+            {:status  200
+             :headers {"Content-Disposition" (str "inline; filename=meeting_" (project-db/meeting-title db id) ".pdf")
```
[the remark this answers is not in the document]
aleksievelizhanin: The meeting header originally entered during meeting creation. PDF generated after that and should not care about sanitazing already created, stored in datomic and read from it meeting title.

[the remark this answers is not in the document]
erno: The newline can't be entered to the meeting title using the browser probably, but you could use the edit and resend xhr function in the browser dev tools console, or alter the meeting title data in the frontend app state atom at the repl. Or could make a backend test that creates a meeting with a newline in the title, and then checks what headers are in the download query response.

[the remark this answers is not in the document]
aleksievelizhanin: What is the point? Is this code unsafe? if so, the whole solution should be re-designed to be entirely on server side.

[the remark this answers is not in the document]
tatut: is the expectation here that the HTTP servers we use are not capable of encoding headers properly? If so, that should be another issue and is not related to meeting PDF.

app/frontend/src/cljs/teet/meeting/meeting_view.cljs
```diff
@@ -892,7 +892,7 @@
         location]
        [(tr "PDF")
         [:a {:href (common-controller/query-url :meeting/download-pdf
-                                                {:db/id (:db/id meeting)})
+                                                {:db/id (:db/id meeting) :meeting meeting})
```
tatut, line 895: Shouldn't pass whole meeting as URL parameters, the payload may be large and the generated URL may be too long. There's already the meeting id, fetch the info on the backend instead of passing whole meeting in URL.

tatut:
LGTM, merge away

app/backend/src/clj/teet/meeting/meeting_queries.clj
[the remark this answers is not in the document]
erno: Yes, something along those lines, https://owasp.org/www-community/attacks/HTTP_Response_Splitting this is the category of vulnerability I suspected. But there isn't any way that newlines can be encoded in HTTP header values AFAIK. Fine by me to create a separate issue for checking this out.

PR #113, opened 2020-12-08
TEET-1069 async processing of THK lambdas

tatut:
Changed lambda processing to be processed in the background (like backup/restore).
The lambda doesn't have to wait for response, as it isn't used.

No tests, as we don't have lambda trigger tests.


outamaa (approved):
Straightforward change, approved.

PR #115, opened 2020-12-10
TEET-1058 Unique metadata

tatut:
Disallow editing a file if the operation would cause two files to have the same metadata.


app/backend/src/clj/teet/file/file_tx.clj
```diff
@@ -67,11 +67,34 @@
...
+                        (d/q '[:find ?f
+                               :where
+                               [?t :task/files ?f]
+                               [(missing? $ ?f :meta/deleted?)]
+                               :in $ ?t]
+                             db task-id))
+        old-file (d/pull db modify-file-keys id)
         new-file (-> file
                      (select-keys modify-file-keys)
                      (update :file/part :db/id)
```
outamaa, line 84: Not part of PR, but what's going on here? Collapsing map containing db id into just the db id itself. Why?

app/backend/src/clj/teet/file/file_tx.clj
```diff
@@ -67,11 +67,34 @@
...
-                     cu/without-nils)]
-    (du/modify-entity-tx old-file new-file)))
+                     cu/without-nils)
+        metadata (file-db/file-metadata (du/entity db id new-file))]
+
+    ;; Check if any other file in this task already has the same metadata
+    ;; that this file would have after applying the tx
+    (if (some #(and (not= % id)
+                      (= metadata
```
outamaa, line 91: indentation

app/backend/test/teet/file/file_commands_test.clj
```diff
@@ -87,3 +96,62 @@
...
+     (thrown-with-msg?
+      Exception #"Two files with same metadata"
+      (tu/local-command :file/modify
+                        {:db/id (:db/id (tu/get-data :second-file))
+                         :file/name "first file.png"
+                         :file/sequence-number 666
+                         :file/document-group (d/pull (tu/db)
+                                                      [:db/id :db/ident]
+                                                      :file.document-group/general)
+                         :file/part nil})))))
```
outamaa, line 157: Good tests!

outamaa (approved):

app/backend/src/clj/teet/file/file_tx.clj
tatut: The map value is what we pull/select of the linked entity, but it isn't valid value for a `ref` attribute. That's why we need to just use the `:db/id`

PR #123, opened 2020-12-14
TEET-1047 Don't include old file versions when checking for task part emptiness

tatut:
Task parts can't be deleted if they contain files.
This check also took into account files that have been replaced and the newer version placed to another part.

Change query to disregard replaced files and add test for this flow.


jkorpelin (approved):
LQTM

PR #125, opened 2020-12-15
TEET-1058 mark2

tatut:
Handle case where retracting attributes didn't check unique properly and add some tests

### DIFF
--- a/app/backend/src/clj/teet/file/file_tx.clj
+++ b/app/backend/src/clj/teet/file/file_tx.clj
@@ -76,20 +76,30 @@
   another file with the same metadata."
   [db {id :db/id :as file}]
   (let [task-id (get-in (du/entity db id) [:task/_files 0 :db/id])
+
+        ;; Take latest versions of all files in task
         task-files (map first
                         (d/q '[:find ?f
                                :where
                                [?t :task/files ?f]
                                [(missing? $ ?f :meta/deleted?)]
+                               (not-join [?f]
+                                         [?replacement :file/previous-version ?f])
                                :in $ ?t]
                              db task-id))
         old-file (d/pull db modify-file-keys id)
-        new-file (-> file
-                     (select-keys modify-file-keys)
-                     ;; Take :db/id from map as the ref attribute value
-                     (update :file/part :db/id)
-                     cu/without-nils)
-        metadata (file-db/file-metadata (du/entity db id new-file))]
+        new-file (select-keys file modify-file-keys)
+
+        ;; Fetch info needed to calculate metadata and
+        ;; make the changes to it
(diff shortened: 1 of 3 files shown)


app/backend/src/clj/teet/file/file_tx.clj
```diff
@@ -76,20 +76,30 @@
...
-        new-file (-> file
-                     (select-keys modify-file-keys)
-                     ;; Take :db/id from map as the ref attribute value
-                     (update :file/part :db/id)
-                     cu/without-nils)
-        metadata (file-db/file-metadata (du/entity db id new-file))]
+        new-file (select-keys file modify-file-keys)
+
+        ;; Fetch info needed to calculate metadata and
+        ;; make the changes to it
```
outamaa, line 94: The changes made could be explained here. Also, `pull-file-metadata-by-id` docstring says it "pull[s] all info that is needed", but here we pull some more. So some context for the extra data would be nice.

app/backend/src/clj/teet/file/file_tx.clj
```diff
@@ -76,20 +76,30 @@
...
   (let [task-id (get-in (du/entity db id) [:task/_files 0 :db/id])
+
+        ;; Take latest versions of all files in task
         task-files (map first
                         (d/q '[:find ?f
                                :where
                                [?t :task/files ?f]
                                [(missing? $ ?f :meta/deleted?)]
+                               (not-join [?f]
+                                         [?replacement :file/previous-version ?f])
```
outamaa, line 87: TIL `not-join`, nice.

app/backend/test/teet/file/file_commands_test.clj
```diff
@@ -152,10 +152,29 @@
...
+                      {:db/id (:db/id (tu/get-data :first-file))
+                       :file/name "first file.png"
+                       :file/document-group nil
+                       :file/sequence-number nil})
+    (let [after (d/pull (tu/db) [:file/document-group :file/sequence-number]
+                        (:db/id (tu/get-data :first-file)))]
+      (is (not (contains? after :file/document-group)) "document group no longer exists")
+      (is (not (contains? after :file/sequence-number)) "seq# no longer exists"))
+
+    (testing "Changing other file to empty as well"
```
outamaa, line 169: 👍

outamaa (approved):
I requested some clarification, otherwise lgtm.

app/backend/src/clj/teet/file/file_tx.clj
tatut: It's the "make changes to it" part. The pull fetches what is already stored in database, but we want the metadata as it would be if we applied the change... that's why we pull the document group info that is sent by client.

PR #128, opened 2020-12-21
Add another integration id migration, as they are still missing in some cases at least in dev env (TEET-947)

erno:


app/backend/src/clj/teet/migration/thk_integration_id.clj
```diff
@@ -23,6 +23,31 @@
...
+        lifecycles (entities-including-attr-and-missing-iid db :thk.lifecycle/id)
+        activities (entities-including-attr-and-missing-iid db :activity/name)
+        tasks (entities-including-attr-and-missing-iid db :task/send-to-thk?)
+        entities (set/union lifecycles activities tasks)]
+    
+    (d/transact
+     conn
+     {:tx-data (vec
+                (for [e entities
+                      :when (nil? (:integration/id e))]
```
tatut, line 47: isn't the `e` here just an id number? not a map that can have `:integration/id`

erno: yep, was left over from old version of code, good catch.

PR #133, opened 2020-12-28
fix situation with conflicting repeatable/non-repeatable migrations & dropped functions

erno:
A function defn (replace_entity_ids) had been removed from the repeatable migrations, and a non-repeatable migration tried to drop it unconditionally. This errored, but changing the nr migration caused a checksum error.

Restore the function in a repeatable migration in placeholder form.


solita-erkkiv (approved):
Changes approved.

PR #135, opened 2021-01-04
TEET-991 Fixes for entity linking - search by file description fix

aleksievelizhanin:
Add latest-only? param to fix searching files by the latest description


app/backend/src/clj/teet/file/file_db.clj
```diff
@@ -213,7 +213,8 @@
                  (assoc latest-version
                    :file/full-name (filename-metadata/metadata->filename
                                      (file-metadata-by-id db (:db/id latest-version)))
-                   :versions previous-versions))))))
+                   :versions (if (not latest-only?) previous-versions)))))))
```
tatut, line 216: Fix lint warning. `if` should always have an else branch, if you want nil as the else, then use `when`

tatut (changes requested):

tatut (approved):

PR #136, opened 2021-01-04
TEET-1061 Task deletion changes

tatut:
- Task can only be deleted if it doesn't have files (it can have deleted files, but not active ones)
- Admin can delete tasks that have separate contract in THK
- Backend tests 
- Simple cypress test for deletion

Also added new local environment route for creating test project/lifecycle/activity/task hiearchy from cypress test. You can call it with `cy.request` from test setup.


app/backend/src/clj/teet/task/task_tx.clj
```diff
@@ -0,0 +1,19 @@
...
+            [teet.meta.meta-model :as meta-model]
+            [datomic.ion :as ion]))
+
+(defn delete-task
+  "Check all preconditions for deleting a task. If precondition
+  fails, cancel the transactions. Otherwise return a tx that marks
+  the task as deleted."
+  [db user task-id]
+  (cond
+    (task-db/task-has-files? db task-id)
```
jkorpelin, line 13: > Manager can see and use the Delete button on Edit task modal only when the task is not marked as "Separate contract (in THK)" and task does not have any current (undeleted files) files. 

Should this AC be checked here also?

tatut: Checked in pre condition of the command, as all authorization stuff is checked at command level, not db

PR #138, opened 2021-01-06
TEET-1148-delete-activities

aleksievelizhanin:
Add Activity TX function to call the check for files that attached to related Tasks


app/backend/src/clj/teet/activity/activity_db.clj
```diff
@@ -181,3 +182,20 @@
        (-> (du/entity db activity-id)
            :activity/name :db/ident)
        #{}))
+
+(defn activity-task-has-files?
+  "Check a task under activity has undeleted files."
+  [db activity-id]
+  (seq
```
jkorpelin, line 189: This returns a list `(true true true)` for example not a boolean. Might be worth it to refactor it so it returns just a boolean.

jkorpelin: Also this whole function could be more readable if it's done using the threading macro 

`(-> tasks (comp :db/id first) .... etc)`

app/backend/src/clj/teet/activity/activity_db.clj
```diff
@@ -181,3 +182,20 @@
...
+    (filter boolean
+      (map (partial file-db/task-has-files? db)
+        (mapv
+          (comp :db/id first)
+          ;; Find all Tasks that belong to Activity
+          (d/q '[:find (pull ?t [:db/id])
```
jkorpelin, line 195: You don't need the `(pull ?t [:db/id])` just having `?t` in it's place returns just the `:db/id`

jkorpelin (commented):
Other than the few minor changes lgtm.

PR #144, opened 2021-01-07
Teet 613 third party opinion

jkorpelin:
Popper tooling and adding of cooperation opinion views for cooperation overview.


tatut (commented):
you should fix unused require lint warning, otherwise looks good

PR #152, opened 2021-01-11
TEET-1193 skip two tasks from being sent to THK

tatut:


app/backend/src/clj/teet/thk/thk_export.clj
```diff
@@ -17,16 +17,26 @@
          :where [?e :thk.project/id _]]
        db))
 
+(def ^:private skip-tasks
+  "Tasks that should never be sent to THK"
+  #{:task.type/design-requirements
+    :task.type/building-permit})
```
outamaa, line 23: This was added in TEET-822:
```
 ;; The tasks below should not be sent to THK
 {:db/ident :migration/TEET-822-retract-thk-task-types
  :txes [[[:db/retract [:db/ident :task.type/design-requirements] :thk/task-type :thk.task-type/design-requirements]
          [:db/retract [:db/ident :task.type/building-permit] :thk/task-type :thk.task-type/building-permit]]]}
```

outamaa: Would it be more robust to fix this by modifying `tasks-to-send` so that the query ensures that the task's `:task/type` has a `:thk/task-type` attribute instead of having a skip set?

outamaa (approved):
If denying these two task types vs checking for existence `:thk/task-type` was what was wanted, lgtm.

PR #153, opened 2021-01-11
Fix upload file version

aleksievelizhanin:


app/backend/src/clj/teet/file/file_db.clj
```diff
@@ -193,27 +193,28 @@
...
-                   :versions (when (seq previous-versions) (previous-versions))))))))
+        (for [f first-versions
+              :let [versions (filter (complement :meta/deleted?)
+                               (reverse
+                                 (take-while some? (iterate next-version f))))
+                    [latest-version & previous-versions] versions]
+              :when latest-version]
+          (when (seq previous-versions)
+                (assoc latest-version
+                  :file/full-name (filename-metadata/metadata->filename
```
jkorpelin, line 215: I think the `file/full-name` is needed in also the cases where previous versions don't exist.

jkorpelin (changes requested):

jkorpelin (approved):
LGTM

PR #159, opened 2021-01-13
Remove some  flakiness from meeting links test

aleksievelizhanin:


aleksievelizhanin:
We have discussed those changes in slack with Joni, so it should be ok to merge.

PR #165, opened 2021-01-13
fix stale url in feedback mailto link (TEET-1184)

erno:
Due to how reagent decides when to re-render a component, the mailto was showing stale info earlier. Fix by passing the url as a parameter to the component, instead of directly reading it from the js environment.


aleksievelizhanin (approved):
Looks legit

PR #166, opened 2021-01-14
TEET-611 application contact info

tatut:


app/backend/src/clj/teet/cooperation/cooperation_commands.clj
```diff
@@ -132,3 +132,40 @@
                 (if (:db/id opinion-form)
                   (meta-model/modification-meta user)
                   (meta-model/creation-meta user)))}]})
+
+(s/def ::contact-form (s/keys :req [:cooperation.contact/name]
```
jkorpelin, line 136: Should this be in cooperation spec to have the same spec for the contact information form?

tatut: Ah yes, I added the form spec later, I'll unify to use the same

jkorpelin (approved):
Other than the small spec suggestion lgtm.

PR #176, opened 2021-01-19
TEET-1191 string length enforcement

tatut:
Add a fallback string length enforcement to all tx-data.

### DIFF
--- a/app/backend/src/clj/teet/db_api/core.clj
+++ b/app/backend/src/clj/teet/db_api/core.clj
@@ -5,7 +5,8 @@
             [teet.log :as log]
             [teet.util.collection :as cu]
             [teet.util.datomic :as du]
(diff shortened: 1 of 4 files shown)


app/backend/src/clj/teet/db_api/core.clj
```diff
@@ -315,6 +316,20 @@
          ~@body))
      (do ~@body)))
 
+(defn- ensure-string-length
+  "Ensure that no strings in tx-data exceed Datomics 4096 character limit."
+  [tx-data]
+  (walk/prewalk
+   (fn [x]
+     (if (and (string? x)
+              (> (.length x) 4096))
```
erno, line 325: use (count x) for more idiomatic Clojure? Or if there's some compelling reason to use the Java method, a comment explaining it.

PR #178, opened 2021-01-19
TEET-1240 fix disabled cypress CI tests

tatut:
cooperation and meeting links tests now work in CI


app/backend/dev-src/teet/test_setup.clj
```diff
@@ -114,10 +122,67 @@
...
+    :link/external-id "2:12601:001:0050",
+    :link/type :cadastral-unit}
+   {:KINNISTU "54128140",
+    :L_AADRESS "Tartu maantee",
+    :TUNNUS "12601:001:0193",
+    :link/external-id "2:12601:001:0193",
+    :link/type :cadastral-unit}])
+
+
+(defn setup-mock-cadastral-unit-link-search [_req]
```
outamaa, line 160: Does this get called anywhere? WIP for now?

tatut: It is used dynamically from the ring handler. Calling `cy.setup(foo, payload)` will invoke `setup-foo` function in the test setup namespace on the backend.

outamaa (approved):

PR #181, opened 2021-01-20
teet-1238 changes format of name to given name - family name in meeti…

jkorpelin:
…ng pdf

change name order in pdf


aleksievelizhanin (approved):
Looks legit

PR #184, opened 2021-01-22
Teet 1205 file description

jkorpelin:
Could especially use some review for the file replacement stuff. Need to create the "full" filename based on the extension of the replacing file and description of the existing file to properly check.


app/common/src/cljc/teet/file/file_model.cljc
```diff
@@ -26,23 +34,40 @@
...
+(defn valid-chars-in-description?
+  [description]
+  (every? allowed-chars description))
+
+(defn valid-description-length?
+  [description]
+  (>= max-description-length (count description)))
+
 (defn validate-file
   "Validate allowed file type and max size. Returns map with error description
```
erno, line 46:
```suggestion
  "Validate allowed file type, legal chars + max length in description and max size. Returns map with error description
```

app/backend/src/clj/teet/file/file_commands.clj
```diff
@@ -257,6 +261,18 @@
    :payload {id :db/id :as file}
    :project-id (project-db/file-project-id db id)
    :authorization {:document/overwrite-document {:db/id id}}
+   :pre [^{:error :description-too-long}
```
erno, line 264: Could these use the validate-file fn?

app/frontend/src/cljs/teet/ui/file_upload.cljs
```diff
@@ -112,52 +109,71 @@
              (not (< 0 sequence-number 10000)))
     {:error :invalid-sequence-number}))
 
+(defn common-file-validation
+  [file]
+  (println "file: " file)
```
erno, line 114:
```suggestion
```

app/frontend/src/cljs/teet/file/file_controller.cljs
```diff
@@ -118,73 +140,96 @@
...
                                                 :file/original-name]))
                      :previous-version-id (:db/id file)}
+           :error-event (partial ->UploadNewVersionError progress-atom)
            :result-event (fn [result]
                            (map->UploadFileUrlReceived
                              (merge result
                                     {:file-data (:file-object new-version)
                                      :on-success (common-controller/->Refresh)})))}))
 
+  UploadNewVersionError
```
erno, line 150: Was there other difference besides the refresh-fx call in this and FileUploadError? Could maybe same code be used for both if the refresh-fx is harmless here too? I don't know what it does so this is shooting in the dark :)

erno (approved):

app/frontend/src/cljs/teet/file/file_controller.cljs
jkorpelin: The other one needs to control atom state of the form, to remove the in-progress state.

PR #188, opened 2021-01-25
Teet 1241 cooperation fixes

jkorpelin:
Adds a separate warning message for the file adding of response when the response already has an opinion and it thus can't be edited.


tatut (approved):

aleksievelizhanin (approved):
Looks legit

PR #195, opened 2021-01-29
Add SSM parameter for notification to environment config

aleksievelizhanin:


app/backend/src/clj/teet/notification/notification_ion.clj
```diff
@@ -6,6 +6,13 @@
             [datomic.client.api :as d]
             [teet.notification.notification-db :as notification-db]))
 
+(defn days-param
+  []
+  (let [param (environment/config-value :notify :application-expire-days)]
+    (if (nil? param)
+      45
+      (Integer/parseInt param))))
```
tatut, line 14: No need to parseInt here or check for nil. You can specify the default value and the function to parse with in the environment ssm definition:
`(->ssm [:notify :application-expire-days] 45 #(Integer/parseInt %))` 

Then the `config-value` will always return an integer directly.

app/backend/src/clj/teet/cooperation/cooperation_db.clj
```diff
@@ -144,6 +144,20 @@
   (get-in (du/entity db application-id)
     [:cooperation.3rd-party/_applications 0 :cooperation.3rd-party/name]))
 
+(defn application-3rd-party-uuid [db application-id]
+  (let [name-of-3rd-party (application-3rd-party db application-id)]
+    (d/q '[:find ?id
```
tatut, line 149: should probably call `first` on result

app/backend/src/clj/teet/cooperation/cooperation_db.clj
```diff
@@ -144,6 +144,20 @@
...
+           :in $ ?name
+           :keys uuid
+           :where [?cooperation :cooperation.3rd-party/name ?name]
```
tatut, line 152: why search 3rd party by name? you can directly navigate `:cooperation.3rd-party/applications` attribute to get the 3rd party

app/backend/src/clj/teet/notification/notification_ion.clj
```diff
@@ -24,11 +25,12 @@
   consultant who entered the application into system
   providing the Activity is not finished"
   ([days]
+   (println (str "days:" days))
```
tatut, line 28: remove debug println

tatut (commented):
may want to get 3rd party directly without querying by name

tatut (approved):

PR #208, opened 2021-02-03
Add dependency to SMTP lib

aleksievelizhanin:
Send emails via SMTP rather than AWS SES

### DIFF
--- a/app/backend/src/clj/teet/environment.clj
+++ b/app/backend/src/clj/teet/environment.clj
@@ -127,7 +127,10 @@
            :kr-subsystem-id (->ssm [:xroad-kr-subsystem-id] nil)}
    :eelis {:wms-url (->ssm [:eelis :wms-url] nil)}
    :email {:from (->ssm [:email :from] nil)
-           :subject-prefix (->ssm [:email :subject-prefix] nil)}
+           :subject-prefix (->ssm [:email :subject-prefix] nil)
+           :host (->ssm [:email :host] nil)
+           :user (->ssm [:email :user] nil)
+           :pass (->ssm [:email :pass] nil)}
    :notify {:application-expire-days (->ssm [:notify :application-expire-days] 45 #(Integer/parseInt %))}})
 
 (defn- load-ssm-config! [base-config]
--- a/app/backend/src/clj/teet/environment.clj
+++ b/app/backend/src/clj/teet/environment.clj
@@ -128,9 +128,7 @@
    :eelis {:wms-url (->ssm [:eelis :wms-url] nil)}
    :email {:from (->ssm [:email :from] nil)
            :subject-prefix (->ssm [:email :subject-prefix] nil)
-           :host (->ssm [:email :host] nil)
-           :user (->ssm [:email :user] nil)
-           :pass (->ssm [:email :pass] nil)}
+           :server (->ssm [:email :server] nil)}
    :notify {:application-expire-days (->ssm [:notify :application-expire-days] 45 #(Integer/parseInt %))}})
 
 (defn- load-ssm-config! [base-config]
--- a/app/backend/src/clj/teet/environment.clj
+++ b/app/backend/src/clj/teet/environment.clj
@@ -129,7 +129,7 @@
    :email {:from (->ssm [:email :from] nil)
            :subject-prefix (->ssm [:email :subject-prefix] nil)
            :contact-address (->ssm [:email :contact-address] nil)
-           :server (->ssm [:email :server] nil)}
+           :server (->ssm [:email :server] {} read-string)}
    :notify {:application-expire-days (->ssm [:notify :application-expire-days] 45 #(Integer/parseInt %))}})
 
 (defn- load-ssm-config! [base-config]
--- a/app/backend/src/clj/teet/integration/integration_email.clj
+++ b/app/backend/src/clj/teet/integration/integration_email.clj
@@ -60,9 +60,11 @@
 (defn send-email-smtp!* [msg]
   (let [smtp-node-config
         (environment/config-map
-          {:host [:email :host]
-           :user [:email :user]
-           :pass [:email :pass]})]
+          {:host [:email :server :host]
+           :user [:email :server :user]
+           :pass [:email :server :pass]
+           :port [:email :server :port]
+           :tls  [:email :server :tls]})]
     (println smtp-node-config)
     (println msg)
     (postal.core/send-message smtp-node-config msg)))
--- a/app/backend/src/clj/teet/integration/integration_email.clj
+++ b/app/backend/src/clj/teet/integration/integration_email.clj
@@ -60,7 +60,7 @@
              "\r\n\r\n"
              (->b64 body))))))
 
-(defn send-email-smtp* [msg]
+(defn send-email-smtp!* [msg]
   (println (str "email server " email-server))
   (println (str "message " msg))
   (println (postal.core/send-message email-server msg)))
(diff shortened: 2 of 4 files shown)


app/backend/src/clj/teet/integration/integration_email.clj
```diff
@@ -57,17 +60,14 @@
...
+  (println (str "message " msg))
+  (println (postal.core/send-message email-server msg)))
 
 (defn send-email! [msg]
   (let [prefix (environment/config-value :email :subject-prefix)
         msg (if prefix
               (update msg :subject #(str prefix " " %))
               msg)]
-    (send-email!* msg)))
+    (send-email-smtp* msg)))
```
tatut, line 73: Take server config from environment in send-email! and pass it to send-email-smtp* function.

Can't use config defined in namespace

tatut: Also should update from master as this has changed there

app/backend/src/clj/teet/integration/integration_email.clj
```diff
@@ -58,13 +61,10 @@
...
-  "Invoke email sending"
-  [msg]
-  (aws/invoke
-   @email
-   {:op :SendRawEmail
-    :request {:RawMessage {:Data (.getBytes (raw-message msg) "UTF-8")}}}))
+(defn send-email-smtp!* [msg]
+  (println (str "email server " email-server))
+  (println (str "message " msg))
+  (println (postal.core/send-message email-server msg)))
```
erno, line 67: I realise this is WIP code but remember to eventually ensure that hostile characters from user entered fields are handled safely or that the postal lib has robust handling for them, to guard against eg header injection.

aleksievelizhanin: the sending emails is automated (mostly). So, the texts are from our code, just user name (which has already sanitized) comes from user. Do you have any particular use case that can be harmful and require additional protection?

app/backend/src/clj/teet/integration/integration_email.clj
```diff
@@ -58,13 +57,13 @@
...
+(defn- with-smtp-config [smtp-conf msg]
+  (assoc msg :server smtp-conf))
+
+(defn send-email-smtp!*
   [msg]
-  (aws/invoke
-   @email
-   {:op :SendRawEmail
-    :request {:RawMessage {:Data (.getBytes (raw-message msg) "UTF-8")}}}))
+  (let [smtp-node-config (:server msg)]
```
tatut, line 65: You should pass server config as separate parameter, not assoc it to the message

app/backend/src/clj/teet/integration/integration_email.clj
```diff
@@ -93,8 +92,15 @@
...
-                 :contact-address [:email :contact-address]})]
-    (send-email!*
+                 :contact-address [:email :contact-address]})
+        smtp-node-config (environment/config-map
+                           {:host [:email :server :host]
+                            :user [:email :server :user]
+                            :pass [:email :server :pass]
+                            :port [:email :server :port]
+                            :tls [:email :server :tls]})]
+    (send-email-smtp!*
```
tatut, line 102: just pass smtp config as first param to `send-email-smtp!*`

app/backend/src/clj/teet/integration/integration_email.clj
```diff
@@ -93,8 +92,15 @@
 (defn send-email! [msg]
   (let [config (environment/config-map
                 {:subject-prefix [:email :subject-prefix]
-                 :contact-address [:email :contact-address]})]
-    (send-email!*
+                 :contact-address [:email :contact-address]})
+        smtp-node-config (environment/config-map
```
tatut, line 96: you could just use `(environment/config-value :email :server)` to get the whole map, no need to separately get each key. That also makes it more future proof as any possible new keys in the server config don't need to be listed here.

app/backend/src/clj/teet/integration/integration_email.clj
erno: I had a quick look and at least subject comes from user supplied data sometimes. We also should be robust against future uses of this code and not rely too much on what we assume is already sanitized and what isn't.

app/backend/src/clj/teet/integration/integration_email.clj
```diff
@@ -3,11 +3,10 @@
   (:require [cognitect.aws.client.api :as aws]
             [clojure.string :as str]
             [teet.environment :as environment]
+            [postal.core :as postal.core]
```
tatut, line 6: no need to use `:as` if no alias is given, you can just require `postal.core` (without the brackets)

app/backend/src/clj/teet/environment.clj
```diff
@@ -128,7 +128,8 @@
    :eelis {:wms-url (->ssm [:eelis :wms-url] nil)}
    :email {:from (->ssm [:email :from] nil)
            :subject-prefix (->ssm [:email :subject-prefix] nil)
-           :contact-address (->ssm [:email :contact-address] nil)}
+           :contact-address (->ssm [:email :contact-address] nil)
+           :server (->ssm [:email :server] nil)}
```
tatut, line 132: We want to parse the string value in SSM to a Clojure map, so the parse-fn should be `read-string` instead of `nil`

app/backend/test/teet/integration/integration_email_test.clj
```diff
@@ -9,7 +9,7 @@
 (defn with-outbox []
   (fn [tests]
     (reset! outbox [])
-    (with-redefs [integration-email/send-email!* #(swap! outbox conj %)]
+    (with-redefs [integration-email/send-email-smtp!* #(swap! outbox conj %1 %2)]
```
tatut, line 12: it would be cleaner if outbox contained maps like `[{:server ... :msg ...}]` instead of alternating server msg.

tatut (changes requested):
Need to check that SSM value is parsed correctly.

app/backend/src/clj/teet/integration/integration_email.clj
aleksievelizhanin: it is sanitized https://github.com/drewr/postal/blob/cfdec2deca126733bd3106792b753c566495685e/src/postal/sendmail.clj#L63

app/backend/test/teet/integration/integration_email_test.clj
```diff
@@ -4,12 +4,14 @@
             [teet.test.utils :as tu]
             [clojure.string :as str]))
 
-(def outbox (atom []))
+(def outbox (atom [:server {} :msg {}]))
```
tatut, line 7: This should be a vector of maps, each containing a `:server` and `:msg` keys.
Now the initial value is a vector of 4 values.

tatut: so initially outbox should be empty `[]`

app/backend/test/teet/integration/integration_email_test.clj
```diff
@@ -4,12 +4,14 @@
...
+
+(defn- update-outbox [server msg] (swap! outbox conj server msg))
```
tatut, line 9: Each message should have server and message, so `(swap! outbox conj {:server server :msg msg})`

app/backend/test/teet/integration/integration_email_test.clj
aleksievelizhanin: Fixed.

app/backend/test/teet/integration/integration_email_test.clj
aleksievelizhanin: Fixed.

tatut (approved):

PR #209, opened 2021-02-04
TEET-1252 add data protection clause to email footer

tatut:


app/backend/src/clj/teet/integration/integration_email.clj
```diff
@@ -65,9 +66,35 @@
...
+
 (defn send-email! [msg]
-  (let [prefix (environment/config-value :email :subject-prefix)
-        msg (if prefix
-              (update msg :subject #(str prefix " " %))
-              msg)]
-    (send-email!* msg)))
+  (-> msg
+      with-subject-prefix
+      with-data-protection-footer
```
outamaa, line 99: Here again, would be better to have a pure function that takes e.g. the email msg and an options map and builds the ready-to-send message. This would make it easy to test is what we intend it to be. I see no need to start refactoring for the sake of it, but maybe something to keep in mind if this gets more complicated in the future.

outamaa (approved):
Some pure functional zealotism aside, good to go.

app/backend/src/clj/teet/integration/integration_email.clj
tatut: Sure, email sending is inherently IO and didn't give it a second thought. Good point.

app/backend/src/clj/teet/integration/integration_email.clj
```diff
@@ -65,9 +66,35 @@
...
+      send-email!*))
```
outamaa, line 100: To put it more clearly, the point was that `send-email!*` is baked in at the end of what's basically pure data transformation pipeline (given that the env accesses were hoisted up from the functions themselves).

PR #214, opened 2021-02-05
TEET-1221 fixes email sending when participants are only users with n…

jkorpelin:
…o email address


aleksievelizhanin (approved):
Looks legit

PR #215, opened 2021-02-05
TEET-983 show new indicator if topic or decision has changed since last viewing

tatut:


app/frontend/src/cljs/teet/meeting/meeting_view.cljs
```diff
@@ -157,53 +166,55 @@
...
   ([param]
    [meeting-view-container param theme-colors/gray-light])
-  ([{:keys [text-color content open? heading heading-button children after-children-component]
+  ([{:keys [text-color content open? heading heading-button children after-children-component on-open]
      :or {text-color :inherit
           open? false}}
     bg-color]
    (r/with-let [open? (r/atom open?)
                 toggle-open! #(do
+                                (when on-open (on-open))
```
jkorpelin, line 175: (on-open) is fired even when closing the view? Is this wanted or should it be just when it's opened?

tatut: good question, I think it is and I should change the name to reflect that.

app/backend/src/clj/teet/meeting/meeting_commands.clj
```diff
@@ -161,10 +162,14 @@
...
-                 :meeting/agenda (mapv #(select-keys % [:db/id
-                                                        :meeting.agenda/topic
-                                                        :meeting.agenda/body
-                                                        :meeting.agenda/responsible])
+                 :meeting/agenda (mapv #(merge
+                                         (select-keys % [:db/id
+                                                         :meeting.agenda/topic
+                                                         :meeting.agenda/body
+                                                         :meeting.agenda/responsible])
+                                         (if (string? (:db/id %))
```
outamaa, line 170: What's the API contract here exactly? If `:db/id` is string, then we're creating a new one, using a placeholder id I guess. Would be nicer to rather communicate what's happening in terms of domain, e.g. `(new-agenda-item? %)` even if the implementation would only be

```
(defn new-agenda-item? [item]
  (string? (:db/id %)))
```

tatut: yes, same as in many places... you can create or update with the same

app/backend/src/clj/teet/meeting/meeting_commands.clj
```diff
@@ -434,3 +439,12 @@
...
           :participation/participant (:db/id participant)})))})
+
+
+(defcommand :meeting/seen
+  {:doc "Mark meeting seen timestamp for user"
+   :payload {id :db/id}
+   :context {:keys [db user]}
+   :project-id (project-db/meeting-project-id db id)
+   :authorization {:project/read-info {:eid (project-db/meeting-project-id db id)}}
+   :transact [(entity-db/entity-seen-tx db user id)]})
```
outamaa, line 450: This reminded me that we discussed whether we should make it a macro expansion time fail if there's no schema for the payload.

tatut: do you mean spec for payload?

app/frontend/src/cljs/teet/meeting/meeting_view.cljs
```diff
@@ -52,6 +52,15 @@
...
 (def milliseconds-when-resent (* 1000 5))                  ;; 5 seconds
 
+(defn- modified-since-last-seen
+  "Show new indicator icon if entity has changed since last seen."
+  [seen-at {:meta/keys [created-at modified-at]}]
+  (let [changed-at (or modified-at created-at)]
+    (when (and seen-at changed-at
+               (> (.getTime changed-at) (.getTime seen-at)))
+      [icons/alert-error-outline {:class "new-indicator"
+                                  :style {:color theme-colors/orange}}])))
```
outamaa, line 62: We now have at least comments and meetings with this kind of functionality. If we get more in the future, it might be worthwhile to come up with a general approach to handling "you haven't seen this / this has changed since you've seen it", both in the application logic and UI level. Just something to keep in mind.

outamaa (approved):

app/backend/src/clj/teet/meeting/meeting_commands.clj
outamaa: Yeah spec. :) Too many years working with Plumatic schema takes it's toll.

tatut: I'll add inline spec for this, but idk if macro expansion time is actually good, as the spec might be registered later.

PR #217, opened 2021-02-08
TEET-1214 delete 3rd party

tatut:


app/backend/src/clj/teet/cooperation/cooperation_commands.clj
```diff
@@ -203,3 +221,13 @@
...
                                {:error :no-contact-info})))})
+
+(defcommand :cooperation/delete-third-party
+  {:doc "Delete a third party. It must not have any applications."
+   :spec (s/keys :req [:db/id])
+   :context {:keys [user db]}
+   :payload {id :db/id}
+   :project-id (cooperation-db/third-party-project-id db id)
+   :authorization {:cooperation/edit-3rd-party {:entity-id id}}
+   :pre [(empty? (:cooperation.3rd-party/applications (du/entity db id)))]
```
jkorpelin, line 233: So you can't delete third parties even if all the applications are marked as deleted?

tatut: oh yes, obviously it should be possible then

PR #222, opened 2021-02-10
TEET-1261 As a user I want to edit and delete cooperation applications

outamaa:
This PR adds functionality for editing and deleting applications. Editing and deleting is only allowed if the application has not been given a third party response.


app/backend/test/cooperation/cooperation_commands_test.clj
```diff
@@ -60,4 +63,69 @@
             (tu/local-command :cooperation/create-application
                               {:thk.project/id project-id
                                :cooperation.3rd-party/name third-party-name
-                               :application invalid-date-application}))))))
+                               :application invalid-date-application}))))
+
+    (testing "Application and project need to match when editing"
+      (is (thrown?
```
tatut, line 69: consider using `is-thrown-with-data?` in test utils that can check for the ex-info

PR #234, opened 2021-02-11
fix format of message for Postal SMTP library

aleksievelizhanin:
Change composing the email message as described for Postal library

### DIFF
--- a/app/backend/src/clj/teet/integration/integration_email.clj
+++ b/app/backend/src/clj/teet/integration/integration_email.clj
@@ -69,18 +69,21 @@
     (update msg :subject #(str prefix " " %))
     msg))
 
+(defn- get-msg-type [msg]
+   (:type (first (:body msg))))
+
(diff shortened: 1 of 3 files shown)


app/backend/src/clj/teet/integration/integration_email.clj
```diff
@@ -69,18 +69,21 @@
...
   "Add data protection clause to text parts"
   [{addr :contact-address} msg]
   (update
-   msg :parts
+   msg :body
    (fn [parts]
      (mapv
       (fn [part]
-        (let [type (get-in part [:headers "Content-Type"])]
+        (let [type (get-msg-type msg)]
```
tatut, line 83: You are processing part but always getting the type from the first part? Shouldn't the type be taken from the current part (just destructure in the `(fn [part] ...`)

tatut (approved):

PR #239, opened 2021-02-12
TEET-1311 clear empty values from cooperation application form

jkorpelin:
Simply remove empty values from the form on send.


aleksievelizhanin (approved):
Looks legit

tatut (approved):

PR #240, opened 2021-02-12
Add trigger event to lambda function

aleksievelizhanin:


app/backend/src/clj/teet/notification/notification_ion.clj
```diff
@@ -4,7 +4,8 @@
             [teet.activity.activity-db :as activity-db]
             [teet.environment :as environment]
             [datomic.client.api :as d]
-            [teet.notification.notification-db :as notification-db]))
+            [teet.notification.notification-db :as notification-db]
+            [clojure.tools.logging :as log]))
```
tatut, line 8: Use `teet.log` as the logging ns like we do everywhere

app/backend/src/clj/teet/notification/notification_ion.clj
```diff
@@ -24,12 +25,13 @@
   new notification is sent to active Activity PM and
   consultant who entered the application into system
   providing the Activity is not finished"
-  ([days]
+  ([event days]
+   (log/info (str "Call notify by event: " event " with days param: " days))
```
tatut, line 29: no need to wrap `(str ...)` to log args

tatut (changes requested):
Logging

tatut (approved):

PR #253, opened 2021-02-16
Filter empty items from the list of tx data

aleksievelizhanin:


app/backend/src/clj/teet/notification/notification_ion.clj
```diff
@@ -35,8 +35,9 @@
   ([event days]
    (log/info "Call notify by event: " event " with days param: " days)
    (let [conn (environment/datomic-connection) db (d/db conn)
-         tx-list (map (comp (partial notify-tx-data db))
-                  (cooperation-db/applications-to-be-expired db days))]
+         tx-list (filter not-empty
+                   (map (comp (partial notify-tx-data db))
```
tatut, line 39: There's nothing wrong with this, but as a note, you can usually use `keep` function (which is like `map` but omits `nil` values) instead of map+filter empty

tatut (approved):

PR #259, opened 2021-02-16
Add more specific error messages for file uploading check. Add where …

aleksievelizhanin:
…condition to search for  No objection coordination or Collaboration type of tasks

### DIFF
--- a/app/backend/src/clj/teet/cooperation/cooperation_db.clj
+++ b/app/backend/src/clj/teet/cooperation/cooperation_db.clj
@@ -152,23 +152,25 @@
                   cooperation-model/application-overview-attrs))]}))
 
 (defn third-party-application-task
-  "Get related task (if any) that has code \"KK\" and can be submitted.
+  "Get the first related task (if any) that has code \"KK\" and can be submitted.
   If no tasks added to the related lifecycle activity return nil
   If there is a task that can not be submitted return map with error message"
   [db third-party-id application-id]
-  (let [tasks (d/q '[:find (pull ?task [*])
-                     :in $ ?third-party-id ?application-id
-                     :where
-                     [?third-part :cooperation.3rd-party/applications ?applications]
-                     [(missing? $ ?applications :meta/deleted?)]
-                     [?applications :cooperation.application/activity ?activities]
-                     [(missing? $ ?activities :meta/deleted?)]
-                     [?activities :activity/tasks ?task]
-                     [(missing? $ ?task :meta/deleted?)]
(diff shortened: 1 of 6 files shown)


app/backend/src/clj/teet/cooperation/cooperation_db.clj
```diff
@@ -152,21 +152,28 @@
...
+                     [?third-part :cooperation.3rd-party/applications ?applications]
+                     [(missing? $ ?applications :meta/deleted?)]
+                     [?applications :cooperation.application/activity ?activities]
+                     [(missing? $ ?activities :meta/deleted?)]
+                     [?activities :activity/tasks ?task]
+                     [(missing? $ ?task :meta/deleted?)]
+                     [?task :task/type ?task-type]
+                     [?task-type :filename/code "KK"]]
+                db third-party-id application-id)
+        can-be-submitted (filter task-model/can-submit? (first tasks))]
```
tatut, line 171: Why the can-be-submitted filter? instead of the previous `ffirst`, in both cases you are only considering the first result from the query.

Seems confusing to call filter on `(first tasks)` as it can only have at most 1 element (the result of pull).

tatut (changes requested):
third-party-application-task should consider all results

app/backend/src/clj/teet/cooperation/cooperation_db.clj
aleksievelizhanin: can-be-submitted filter? is used to filter out the tasks that can not be submitted. And have a possibility to compare the list of tasks including ones that can not be submitted and the filtered result. It enables showing the error to UI about the task in the non-submittable state. 
Yes, (first tasks) is not obvious until I wrote the test and realised that (filter) doesn't filter the [[task]], because it is vector of vectors. So, I have to call (first) to have [task] and apply filter to it.
This function can at max return only one task or no tasks at all. Hope it explains the purpose of the changes.

If there is a use case then this function will fail I am eager to fix it.

tatut: it doesn't "filter out the tasks" it only checks the 1st returned task from the query

tatut: so if the query returns 2 tasks, and the first one is not submittable but the second one is... this would fail as it only considers the submittability of the first one

aleksievelizhanin: Fixed.

tatut (approved):

PR #271, opened 2021-02-19
TEET-1307 - Improved schema fix for design road safety task thk type

erno:


app/backend/resources/schema.edn
```diff
@@ -1737,4 +1737,11 @@
...
            :db/ident :entity-seen/entity+user}
           {:db/id :file-seen/seen-at
-           :db/ident :entity-seen/seen-at}]]}]]}]
+           :db/ident :entity-seen/seen-at}]]}
+ {:db/ident :migration/road-safety-design-and-construction-codes-v3
+  :txes [[{:db/ident :thk.task-type/design-road-safety-audit :thk/code "4011"}]
+         [{:db/ident :task.type/design-road-safety-audit,
+           :enum/attribute :task/type,
+           :enum/valid-for :task.group/design-approval,
+           :thk/task-type :thk.task-type/design-road-safety-audit,
```
jkorpelin, line 1746: Should all of this non-changed stuff be in the transaction? If we are only changing the :thk/task-type value?

erno: Good point. The reason for including the redundant stuff was to make the schema more readable, but yep should make sure the reader know what's changed.

I converted the redundant fields to comments now and documented that's changed. Though still there are advantages of keeping the redundant attributes as real schema items instead of comments, because comments can be wrong :) But changed per your suggestion to keep the discussion brief.

PR #281, opened 2021-02-24
TEET-1336 delete cooperation application response

tatut:


app/frontend/src/cljs/teet/cooperation/cooperation_view.cljs
```diff
@@ -662,6 +662,19 @@
                                   (fn [e!]
                                     (e! (close-event))
                                     (e! (cooperation-controller/->ResponseCreated response (boolean (:db/id @form-atom)))))))
+                 :delete (when (:db/id @form-atom)
```
jkorpelin, line 665: Does this need some authorization checking? Or should the defcommands pre check be used here to show a disabled delete button when an opinion exists?

tatut: If you can create/edit, then you can also delete afaict

jkorpelin (approved):
If the situations in the comment are taken care of somewhere else this looks good.

app/frontend/src/cljs/teet/cooperation/cooperation_view.cljs
tatut: The pre check is already checked, the "edit" button isn't shown if the application has an opinion

PR #288, opened 2021-02-25
translation fixes

kristjanZ:


kristjanZ:
actions failed. probably someone else will update the translations anyway.

PR #293, opened 2021-02-26
TEET-1349 fixes cooperation application activity matching

jkorpelin:
Previously it could find activities that are deleted.

### DIFF
--- a/app/backend/src/clj/teet/cooperation/cooperation_db.clj
+++ b/app/backend/src/clj/teet/cooperation/cooperation_db.clj
@@ -179,16 +179,20 @@
   "Given project-id and an application with a date, return an activities id that is on going during the dates"
   [db project-id application]
   (let [application-date (:cooperation.application/date application)]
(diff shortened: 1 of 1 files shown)


app/backend/src/clj/teet/cooperation/cooperation_db.clj
```diff
@@ -179,16 +179,20 @@
...
-                            :activity/name
-                            :activity/estimated-start-date
-                            :activity/estimated-end-date]}]}]
-                   [:thk.project/id project-id]))
-         (mapcat
-           :thk.lifecycle/activities)
+    (->> (mapv
+           first
+           (d/q '[:find (pull ?ac [:db/id
+                                   :meta/deleted?
```
tatut, line 185: why pull `:meta/deleted?` if you are checking that it is missing below... it can't have value.

jkorpelin: True it was left over from debugging.

PR #297, opened 2021-02-26
TEET-1270 When deleting user supplied files and attachments, delete corresponding models from Vektorio

erno:


app/backend/src/clj/teet/file/file_commands.clj
```diff
@@ -81,18 +83,30 @@
...
+   ;; will do a permission check depending on the attachment type.
    :pre [(or attached-to
              (file-db/own-file? db user file-id))]
-   :transact (if attached-to
-               (file-db/delete-attachment db user file-id attached-to)
-               [(deletion-tx user file-id)])})
+   :transact (let [file-delete-tx (if attached-to
+                                    (file-db/delete-attachment db user file-id attached-to)
+                                    [(deletion-tx user file-id)])]
+               (maybe-vektorio-delete! db file-id)
```
tatut, line 108: can attachments be models? or should they only be proper task files

app/backend/src/clj/teet/file/file_commands.clj
```diff
@@ -241,9 +255,11 @@
    :payload {:keys [file-id]}
    :project-id (project-db/file-project-id db file-id)
    :authorization {:document/delete-document {:db/id file-id}}
-   :transact (vec
-              (for [version-id (file-db/file-versions db file-id)]
-                (deletion-tx user version-id)))})
+   :transact (let [file-delete-tx (vec
+                              (for [version-id (file-db/file-versions db file-id)]
+                                (deletion-tx user version-id)))]
+               (maybe-vektorio-delete! db file-id)
```
tatut, line 261: should all versions be deleted? or do later versions update the same model in vektorio?

erno: I've understood that we update the same model in vektorio.

app/backend/src/clj/teet/file/file_commands.clj
erno: Good question. The delete code can check for existence of the vektorio attributes. I actually had a bug there since I hadn't tested with files that don't have vektorio models, i fixed that now in vektorio-core/delete-file-from-project! fn

PR #300, opened 2021-03-02
TEET-1269 scheduled Vektorio (re)upload

erno:


solita-erkkiv (approved):

app/backend/src/clj/teet/file/file_db.clj
```diff
@@ -451,6 +453,27 @@
           :in $ ?uuid]
         db uuid)))
 
+;; used by vektorio scheduled upload retry. considering only files that are older than threshold,
+;; so we don't step on top of normally (by s3 trigger) started uploads.
+(defn recent-task-files-without-model-id [db threshold-in-minutes]
```
jkorpelin, line 458: Should this also check that the files belong to a not-deleted task/activity?

erno: Yep, I guess this PR is a fine place to handle this as well. I added these checks to clojure code using a separate file-db/is-task-file? fn now to both normal and scheduled import code paths.

PR #304, opened 2021-03-02
Add ion for removing all user email addresses

erno:


app/backend/src/clj/teet/backup/backup_ion.clj
```diff
@@ -354,3 +354,17 @@
...
+(defn delete-user-email-addresses [_]
+  (let [db-connection (environment/datomic-connection)
+        db (d/db db-connection)
+        users-result (d/q '[:find ?u ?email :keys user-eid email
+                            :where [?u :user/email ?email]]
+                          db)
+        retract-tx (vec (for [{:keys [user-eid email]} users-result]
+                          [:db/retract user-eid :user/email email]))]    
+    (d/transact db-connection {:tx-data retract-tx})
+    (log/info "Removed email address from" (count retract-tx) "users")))
```
tatut, line 370: should you return the string json success as we do elsewhere so that lambda doesn't think it failed and retry it

erno: good catch.

PR #311, opened 2021-03-03
TEET-775 Fix exchange property text parameters

solita-erkkiv:


app/frontend/src/cljs/teet/ui/typography.cljs
```diff
@@ -56,6 +56,14 @@
...
    :textTransform :uppercase})
 
+(defn- sub-section-style []
+  {:font-weight    700
+   :font-size      "1rem"
+   :margin-bottom "0.5rem"
+   :margin-top "1.5rem"
+   :color         theme-colors/secondary-text
+   :text-transform :uppercase})
+
```
aleksievelizhanin, line 66: No need for extra empty line here?

app/frontend/src/cljs/teet/ui/typography.cljs
```diff
@@ -88,6 +96,8 @@
 
 (def DataLabel (util/make-component :div {:class (<class data-label-style)}))
 
+(def SubSectionLabel (util/make-component :div {:class (<class sub-section-style)}))
```
jkorpelin, line 99: This could be a span or p since div's shouldn't contain text without a wrapping text element. Other than that looks ok

PR #313, opened 2021-03-04
Teet 1255 vektorio dwg isssues

jkorpelin:


solita-erkkiv (approved):

app/backend/src/clj/teet/file/file_import.clj
```diff
@@ -27,14 +27,17 @@
...
   (let [suffix (uploaded-file-suffix ctx)]
     (when-let [extensions (and vektorio-config (get-in vektorio-config [:config :file-extensions]))]
       (when (extensions suffix)
         (let [db (d/db conn)
               file-id (ffirst (d/q '[:find ?file
                                      :in $ ?s3-key
-                                     :where [?file :file/s3-key ?s3-key]]
-                                   (:file-key s3)))]
+                                     :where
+                                     [_ :task/files ?file]  ;; Check that file belongs to task so we don't upload attachments
```
erno, line 38: this is now redundant with the is-task-file? check below

jkorpelin: The is-task-file? seems to be doing a lot of stuff compared to just having the where clause, but you're correct I didn't notice that check was already added added.

PR #320, opened 2021-03-05
Do check for project id and task first in uploading file validation

aleksievelizhanin:
Change the order of validations for uploaded files to catch files with wrong metadata and illigal characters in file name.


app/frontend/src/cljs/teet/ui/file_upload.cljs
```diff
@@ -141,38 +141,38 @@
...
-        (when-let [error (or (and metadata (validate-name file-row))
-                             (validate-seq-number file-row)
-                             (file-model/validate-file file-info)
-                             (file-model/validate-file-metadata project-id task metadata))]
+  (let [file-info (files-field-entry file-row)
+        file-row-info (merge file-row file-info)]
+    (or (common-file-validation file-row-info)
+      (when-let [error (or (and metadata (validate-name file-row-info))
+                          (validate-seq-number file-row-info)
+                          (file-model/validate-file file-row-info)
```
outamaa, line 149: 1. Nitpicking: Indentation of `when-let`.

2. If I parse this correctly, we now merge `file-info` and `file-row` and pass that to all the validation functions whereas earlier it was either or. Where does the wanted change come from?

3. `common-file-validation` calls `file-model/validate-file` internally, so at least now with merged data it seems to me the call on line 149 is unnecessary.

aleksievelizhanin: 1. fixed.
2. because this was validated in the commoon-file-validation and if not merged this changes was not visible for the rest of validations.
3. redundant call is removed

outamaa (approved):

PR #324, opened 2021-03-08
Add missing vektorio/project-id to projects list of attributes, so it…

aleksievelizhanin:
Add :vektorio/project-id so it will be available to BIM link logic
Call add user to project then project is created


aleksievelizhanin:
@jkorpelin looks legit to me

PR #326, opened 2021-03-09
Fix HTTP 413 too big payload by removing whole project from input par…

aleksievelizhanin:
…ameters

### DIFF
--- a/app/frontend/src/cljs/teet/project/project_navigator_view.cljs
+++ b/app/frontend/src/cljs/teet/project/project_navigator_view.cljs
@@ -433,7 +433,9 @@
         (when (and
                 (common-controller/feature-enabled? :vektorio)
                 (some? (:vektorio/project-id project)))
-          [common/vektorio-link {:href (common-controller/query-url :vektorio/instant-login project)
+          [common/vektorio-link {:href (common-controller/query-url
+                                         :vektorio/instant-login {:db/id (:db/id project)
+                                                                  :vektorio/project-id (:vektorio/project-id project)})
                                  :target "_blank"}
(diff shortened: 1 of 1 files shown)


app/frontend/src/cljs/teet/project/project_navigator_view.cljs
```diff
@@ -433,7 +433,9 @@
         (when (and
                 (common-controller/feature-enabled? :vektorio)
                 (some? (:vektorio/project-id project)))
-          [common/vektorio-link {:href (common-controller/query-url :vektorio/instant-login project)
+          [common/vektorio-link {:href (common-controller/query-url
+                                         :vektorio/instant-login {:db/id (:db/id project)
```
jkorpelin, line 437: could be done with (common-controller/query-url :vektorio/instant-login (select-keys project [:db/id :vektorio/project.id])) as well.

app/frontend/src/cljs/teet/project/project_navigator_view.cljs
```diff
@@ -433,7 +433,9 @@
...
+                                                                  :vektorio/project-id (:vektorio/project-id project)})
                                  :target "_blank"}
            (str (tr [:project :bim-models]) (:vektorio/project-id project))])
```
jkorpelin, line 440: There should probably be a space between the translation and the vektorio project-id.

jkorpelin (approved):

app/frontend/src/cljs/teet/project/project_navigator_view.cljs
aleksievelizhanin: There is no space between THK and project id, but I agree with a space it looks better, just different from THK link

PR #328, opened 2021-03-09
TEET-1256 default project creation to use thk-id if no project name e…

jkorpelin:
…xists.

### DIFF
--- a/app/backend/src/clj/teet/integration/vektorio/vektorio_core.clj
+++ b/app/backend/src/clj/teet/integration/vektorio/vektorio_core.clj
@@ -21,9 +21,11 @@
 (defn create-project-in-vektorio!
   [conn vektor-config project-eid]
   (let [db (d/db conn)
-        project (d/pull db [:db/id :thk.project/name :thk.project/project-name] project-eid)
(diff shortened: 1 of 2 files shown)


erno (approved):

app/backend/src/clj/teet/integration/vektorio/vektorio_core.clj
```diff
@@ -21,9 +21,11 @@
 (defn create-project-in-vektorio!
   [conn vektor-config project-eid]
   (let [db (d/db conn)
-        project (d/pull db [:db/id :thk.project/name :thk.project/project-name] project-eid)
+        project (d/pull db [:db/id :thk.project/name :thk.project/project-name :thk.project/id] project-eid)
         project-name-for-vektor (or (:thk.project/project-name project)
-                                    (:thk.project/name project))
+                                    (:thk.project/name project)
+                                    (:thk.project/id project) ;; default to thk-id of the project, because might have no name in dev
+                                    )
```
aleksievelizhanin, line 28: I would get rid of comment and dangling braket by one additional binding with "default-vektor-project-name" 
So, the code could be self-explaining.

PR #329, opened 2021-03-09
TEET-1255 Fixes project name for vektorio

jkorpelin:
Fixes project name for vektor to contain the thk id as well as the project name


aleksievelizhanin (approved):
Looks legit

PR #334, opened 2021-03-10
TEET-1393 ROTL update

kristjanZ:
Raul asked for the update


app/backend/resources/asset-schema.edn
```diff
@@ -258,1388 +266,1412 @@
...
+  :asset-schema/label [true "Trepp"],
+  :asset-schema/description ["Stair" ""],
   :ctype/parent ":ctype/bridgesupport"}
  {:db/id ":ctype/expansionjointsupportbeam",
   :db/ident :ctype/expansionjointsupportbeam,
-  :db/doc "No",
   :asset-schema/type :asset-schema.type/ctype,
-  :asset-schema/label
-  ["Paisumisvuugi tugipruss" "Expansion joint support "],
+  :asset-schema/label [true "Paisumisvuugi tugipruss"],
```
jkorpelin, line 485: Can't have true in place of the label as it's defined as tuple fo strings. ["string1" "string2"]

kristjanZ: image
aa okey so it's because Raul has added column C. 
I'll try to work a round it.

jkorpelin (approved):

PR #340, opened 2021-03-15
Teet 1356 add opinion of land owner

jkorpelin:
Most new stuff is related to land owner opinions which is hidden behind a feature flag in the frontend. Also refactored land section views to a better place and moved styles to a new ns.


aleksievelizhanin (approved):
Looks legit

PR #344, opened 2021-03-16
TEET-1255 uses clj-http instead of httpkit in the vektorio posting to…

jkorpelin:
… fix large file uploads

### DIFF
--- a/app/backend/src/clj/teet/integration/vektorio/vektorio_client.clj
+++ b/app/backend/src/clj/teet/integration/vektorio/vektorio_client.clj
@@ -44,7 +44,8 @@
                                           payload
                                           (cheshire/encode payload))})
                   (catch Exception e
-                    (log/fatal e "Exception in vektorio post for url " (str api-url endpoint))))]
+                    (log/fatal e "Exception in vektorio post for url " (str api-url endpoint))
+                    (throw e)))]
     (vektor-message-handler resp)))
 
 (defn vektor-delete!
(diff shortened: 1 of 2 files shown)


app/backend/src/clj/teet/integration/vektorio/vektorio_client.clj
```diff
@@ -37,14 +38,15 @@
...
-                            :body (if (= content-type "application/octet-stream")
-                                    payload
-                                    (cheshire/encode payload))})]
+        resp (try (clj-http/post (str api-url endpoint)
+                                 {:headers headers
+                                  :body (if (= content-type "application/octet-stream")
+                                          payload
+                                          (cheshire/encode payload))})
+                  (catch Exception e
+                    (log/fatal e "Exception in vektorio post for url " (str api-url endpoint))))]
```
tatut, line 47: what will `vektor-message-handler` do in the catch case, resp will be `nil`. Should the exception be rethrown to stop it from trying to handle nil

app/backend/src/clj/teet/integration/vektorio/vektorio_client.clj
```diff
@@ -37,14 +38,16 @@
                         "Content-Type" "application/json"}
                        headers)
         content-type (get headers "Content-Type")
-        resp @(http/post (str api-url endpoint)
-                           {:headers headers
-                            :body (if (= content-type "application/octet-stream")
-                                    payload
-                                    (cheshire/encode payload))})]
+        resp (try (clj-http/post (str api-url endpoint)
```
erno, line 41: A comment about why/how clj-http fixes things for us would be good here

PR #353, opened 2021-03-19
TEET-1398 hack for opening main menu

kristjanZ:
This seemed to solve the problem when running it locally. Maybe there's a better option to open the main menu?

### DIFF
--- a/ci/browser-tests/cypress-localdev.json
+++ b/ci/browser-tests/cypress-localdev.json
@@ -1,3 +1,3 @@
 {"baseUrl": "http://localhost:4000/",
-"viewportWidth":  1600,
-"viewportHeight":  1024}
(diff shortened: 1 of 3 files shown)


jkorpelin (approved):
Tested it also, looks good for now. Probably should remove the estonian language from the other parts of the tests, and maybe use an id for the left navigation menu  to search by instead of the contains("menu").

kristjanZ:
Just to be clear in order to use the ID it should first be created or it already exists and I missed it?
Also related to may be the viewports. Should we unify viewportWidth and viewportHeight in cypress-ci.json and cypress-localdev.json or there is some point to them being different?

jkorpelin:
It should be added, but don't think it's necessary now. There's no reason for the width and height to be different in the different configs.

PR #368, opened 2021-03-24
Teet 1366 show estate names in html

jkorpelin:
Adds estate names to the estate rows if they exist, and cypress test for land owner opinion export


aleksievelizhanin (approved):
Looks good to me

PR #370, opened 2021-03-25
Profiling tools for local dev

outamaa:
This PR adds profiling tools for use in local dev. We don't necessarily need to merge this, but might be useful.


tatut (approved):

tatut:
```Caused by: java.io.FileNotFoundException: Could not locate clj_async_profiler/core__init.class, clj_async_profiler/core.clj or clj_async_profiler/core.cljc on classpath. Please check that namespaces with dashes use underscores in the Clojure file name.
27
```

the cypress test profile can't load user namespace

PR #371, opened 2021-03-25
TEET-1454 remove un wanted status of opinion and response from html

jkorpelin:
removes not wanted attributes from the export html


aleksievelizhanin (approved):
Looks legit

PR #376, opened 2021-03-26
Teet 1461 duplicate file upload

jkorpelin:
fixes duplicate file uploading


tatut (approved):

aleksievelizhanin (approved):
Looks legit. Some strange constants (666, for example), but for tests it is probably ok.

PR #381, opened 2021-03-29
TEET-1357 and TEET-1358 fixes

aleksievelizhanin:
- opinion author relationship with land is not in bold style as designed
- if the relationship empty the layout is still the same (place is taken even if empty)
- opinions list rows aligned to the left as designed
- update opinions count after delete

### DIFF
--- a/app/frontend/src/cljs/teet/land/owner_opinion_view.cljs
+++ b/app/frontend/src/cljs/teet/land/owner_opinion_view.cljs
@@ -273,6 +273,7 @@
    (doall
      (for [[label data is-bold is-label-hidden :as row] data
            :when row]
+       ^{:key (str data "-" label)}
        [:div
         {:class [(<class common/info-row-item-style false)]}
         (if (true? is-label-hidden)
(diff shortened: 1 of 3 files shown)


app/frontend/src/cljs/teet/land/owner_opinion_view.cljs
```diff
@@ -266,14 +266,38 @@
...
+            [typography/Text label]))
+        [:div data]]))])
+
 (defn owner-opinion-heading [opinion]
-  [common/basic-information-row
-   [[(:land-owner-opinion/respondent-name opinion) (get-activity-name opinion)]
-    [(:land-owner-opinion/respondent-connection-to-land opinion) (get-activity-type opinion)]]])
+  [opinion-list-header-row
+   [[(:land-owner-opinion/respondent-name opinion) (get-activity-name opinion) true false]
+    ["." "/" false true]
```
jkorpelin, line 294: These being a map with keys explaining what the values are could be much more readable.

jkorpelin (approved):
other than the one comment looks good to me.

app/frontend/src/cljs/teet/land/owner_opinion_view.cljs
aleksievelizhanin: Refactored

PR #385, opened 2021-03-30
TEET-1480 Restrict the characters allowed to be entered to task part …

solita-erkkiv:
…names


jkorpelin:
You could also add the same validation to the command as well.

tatut (approved):

PR #392, opened 2021-03-31
Add multimethod to provide compare functions for complex column data

aleksievelizhanin:


app/common/src/cljc/teet/project/project_model.cljc
```diff
@@ -92,6 +92,15 @@
 
 (defmulti get-column (fn [_project column] column))
 
+(defmulti get-column-compare (fn [column] column))
+
+(defmethod get-column-compare :default [_] nil)
+
+(defmethod get-column-compare :thk.project/activity-status [_]
+  (fn [x y] (compare
```
tatut, line 100: I don't understand what is the need for this new multi method.

And why are we comparing two keyword idents together? Is there a sort order for the keywords or should they be compared based on the translated labels?

aleksievelizhanin: The need for the multimethod is to provide compare for the particular column (or columns in general)
How it should be sorted is not specified in the issue. Will ask ...

app/frontend/src/cljs/teet/ui/table.cljs
```diff
@@ -88,37 +88,41 @@
...
+                              :filters filters
+                              :columns columns
+                              :filter-type filter-type}]
+             [TableBody {}
+              (doall
+                (for [row (take @show-count
+                            ((if (= sort-dir :asc) identity reverse)
+                             (if get-column-compare
+                               (sort-by #(get-column % sort-col)
+                                 (fn [x y]
```
tatut, line 114: Doesn't the `get-column-compare` take one argument (a column) and return a comparison function?
apply is calling it here with 2 values (x and y)

aleksievelizhanin: Right, fixed.

app/frontend/src/cljs/teet/ui/table.cljs
```diff
@@ -88,37 +88,41 @@
...
+                              :columns columns
+                              :filter-type filter-type}]
+             [TableBody {}
+              (doall
+                (for [row (take @show-count
+                            ((if (= sort-dir :asc) identity reverse)
+                             (if (get-column-compare sort-col)
+                               (sort-by #(get-column % sort-col)
+                                 (fn [x y]
+                                   (apply (get-column-compare sort-col) [x y])) data)
```
tatut, line 115: no need to use apply, `((get-column-compare sort-col) x y)` works as well

app/frontend/src/cljs/teet/ui/table.cljs
```diff
@@ -88,37 +88,40 @@
...
+                              :sort-column sort-col
+                              :sort-direction sort-dir
+                              :filters filters
+                              :columns columns
+                              :filter-type filter-type}]
+             [TableBody {}
+              (doall
+                (for [row (take @show-count
+                            ((if (= sort-dir :asc) identity reverse)
+                             (if (get-column-compare sort-col)
```
tatut, line 112: could still be more clearly formatted, like:
```clojure
(sort-by #(get-column % sort-col) 
         (or (get-column-compare sort-col) compare)
         data)
```

tatut: shouldn't need `if` or wrapping anon function, but functionally that is ok

erno (approved):

PR #397, opened 2021-04-06
TEET-1509 don't store THK update timestamps

tatut:
THK seems to change update timestamps every time it imports CSV from TEET. This causes every object to be "changed" when they send it back and causes unnecessary transaction growth on TEET side.

Don't store the THK update timestamps in integration info (send them back empty) as THK doesn't seem to use them.
Also sorts the integration info by keys so it is stable.


solita-erkkiv (approved):

aleksievelizhanin:
Looks legit

PR #399, opened 2021-04-06
TEET-1482 Unable to add cadastral units/restrictions by drawing a selection on the map

outamaa:
This PR fixes a bug which made it impossible to select cadastral units and restrictions by drawing the selected are on the map.

The fix replaces the use of `reagent.core/wrap` with the use of `reagent.core/cursor` and `reagent.ratom/make-reaction` to provide `panels/modal` with an atom-like object that behaves in the way we want.

One downside of the fix here is that `draw-selection` component now directly accesses the app state by constructing the cursor. However, all the events used from the `teet.project.project-controller` do the same.

### DIFF
--- a/app/frontend/src/cljs/teet/ui/panels.cljs
+++ b/app/frontend/src/cljs/teet/ui/panels.cljs
@@ -141,7 +141,7 @@
                              [icons/navigation-close]]]
     [Dialog {:max-width max-width
              :full-width true
-             :open @open-atom
+             :open (boolean @open-atom)
              :on-close close-fn
              :data-cy data-cy}
      (if title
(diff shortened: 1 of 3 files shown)


app/frontend/src/cljs/teet/ui/panels.cljs
```diff
@@ -141,7 +141,7 @@
                              [icons/navigation-close]]]
     [Dialog {:max-width max-width
              :full-width true
-             :open @open-atom
+             :open (boolean @open-atom)
```
outamaa, line 144: A minor tweak to enable nil punning.

jkorpelin (approved):

PR #401, opened 2021-04-06
TEET-1440 changes markdown parsing library, so inline styles don't break

jkorpelin:
This changes the markdown parsing library to this: https://github.com/Rosey/markdown-draft-js

Which fixes the inline styles being broken when they end in empty space.


app/frontend/package.json
```diff
@@ -14,6 +14,7 @@
     "draft-js-import-markdown": "^1.4.0",
```
tatut, line 14: should the old library be removed?

jkorpelin: Yea, true good idea.

tatut (approved):

PR #405, opened 2021-04-07
TEET-1451 Enable user-based access to Vektor

solita-erkkiv:


app/backend/src/clj/teet/integration/vektorio/vektorio_client.clj
```diff
@@ -89,12 +99,18 @@
...
+  [conn user vektor-conf project-id user-id]
   (assert (some? project-id) "Must specify the project to which the user is added")
   (assert (some? user-id) "Must specify the user which is added to the project")
   (log/info "Adding vektorio user" user-id "to vektorio project:" project-id)
-  (vektor-post! vektor-conf {:endpoint (str "projects/" project-id "/users")
-                             :payload {:userId user-id}}))
+  (when
+    (empty?
+      (d/q '[:find ?pid :where [?u :vektorio/granted-projects ?pid] :in $ ?u ?pid]
+         (d/db conn) (user-model/user-ref user) project-id))
```
outamaa, line 109: This addition introduces a dependency to datomic. Let's keep this namespace purely a wrapper for the Vektorio HTTP API and instead pass make this check in the namespace where `add-user-to-project!` is called.

app/backend/src/clj/teet/integration/vektorio/vektorio_client.clj
```diff
@@ -89,12 +99,18 @@
...
+      (vektor-post! vektor-conf {:endpoint (str "projects/" project-id "/users")
+                             :payload {:userId user-id}})
+             (d/transact conn {:tx-data [{:db/id (:db/id user)
+                                          :vektorio/granted-projects project-id}]})))
```
outamaa, line 113: Continuing with the Datomic dependency, if the addition is successful, the calling code can transact the project id to Datomic. This function should only be a wrapper for the HTTP request.

outamaa (commented):
The Datomic dependent code should be hoisted from `vektorio-client` to `vektorio-core` as the former should just be a wrapper for the HTTP API with minimal dependencies. Otherwise looks good to me.

PR #406, opened 2021-04-07
TEET-1484 localization and validation fix

tatut:


app/backend/src/clj/teet/asset/asset_commands.clj
```diff
@@ -21,11 +21,14 @@
...
    :transact
-   ^{:db :asset}
-   [(merge {:asset/project project-id}
-           (asset-type-library/form->db
-            (asset-type-library/rotl-map (asset-db/asset-type-library adb))
-            asset))]})
+   (with-meta
+     (du/modify-entity-retract-nils
+      adb
+      (merge {:asset/project project-id}
```
outamaa, line 27: Do we check that the asset belongs to the project provided as argument? Do we need to?

app/backend/src/clj/teet/asset/asset_commands.clj
```diff
@@ -49,12 +52,16 @@
...
+       adb
+       (asset-type-library/form->db
+        (asset-type-library/rotl-map (asset-db/asset-type-library adb))
+        (dissoc component :component/components)))
+      ;; Link this to parent
+      [:db/add parent-id
+       (case (asset-db/item-type adb parent-id)
+         :asset :asset/components
+         :component :component/components) (:db/id component)])
+     {:db :asset})})
```
outamaa, line 67: We could maybe add `:db` as an optional key in the map provided to `defquery`?

app/backend/src/clj/teet/asset/asset_commands.clj
tatut: you mean that you could modify an existing asset and change it to the current project? Probably should check

app/common/src/cljc/teet/util/datomic.cljc
```diff
@@ -199,6 +199,24 @@
...
+     (let [entity-values (cu/without-nils new-entity)
+           nil-attributes (into []
+                                (comp
+                                 (filter (comp nil? val))
+                                 (map key))
+                                new-entity)]
+       (into [entity-values]
+             (when (number? id)
+               (for [[attr val] (d/pull db nil-attributes id)]
+                 [:db/retract id attr val]))))))
```
outamaa, line 218: As an aside, this could be used to refactor lots of our backend code to something like
```
(-> (pull-entity-from-db entity-id)
    (do-domain-operation other-request-args)
    (persist-using-modify-entity-retract-nils-and-meta-stuff user)
```
where `do-domain-operation` could then be more easily tested in isolation.

outamaa (approved):

app/backend/src/clj/teet/asset/asset_commands.clj
outamaa: Yes, I assumed this can be used to also modify an existing cost item.

PR #414, opened 2021-04-09
Change the date of Cooperation application from datepicker to label w…

aleksievelizhanin:
…ith a tooltip


app/frontend/src/cljs/teet/cooperation/cooperation_view.cljs
```diff
@@ -953,6 +953,22 @@
...
+(defn format-date
+  "Format date to dd.MM.yyyy"
+  [date]
+  (cljs-time.format/unparse custom-date-formatter date))
+
+(defn date-label-component
+  [{value :value label :label}]
+  [:div {:style (merge {:color theme-colors/gray} common-styles/body-2-bold)} label
+   [:div {:style common-styles/body-1-bold}
+    [common/popper-tooltip {:title "Date of application can not be changed"
```
tatut, line 968: English text here? should be translated

jkorpelin, line 968: Should the title be a translation?

tatut (changes requested):
translate text

app/frontend/src/cljs/teet/cooperation/cooperation_view.cljs
```diff
@@ -953,6 +953,22 @@
                                     :data-cy "add-contact"}
             (tr [:cooperation :add-application-contact])])]])]))
 
+(def custom-date-formatter
+  (cljs-time.format/formatter "dd.MM.yyyy"))
+
+(defn format-date
+  "Format date to dd.MM.yyyy"
+  [date]
+  (cljs-time.format/unparse custom-date-formatter date))
```
jkorpelin, line 962: this date formatting should probably be in the `teet.util.date.cljc`

jkorpelin: or use the `teet.ui.format/date` fn

app/frontend/src/cljs/teet/cooperation/cooperation_view.cljs
aleksievelizhanin: Yes, waiting the verbiage from design..

aleksievelizhanin: Yes, waiting for the exact verbiage from design..

app/frontend/src/cljs/teet/ui/common.cljs
```diff
@@ -592,3 +593,19 @@
...
+  "Shows :label with formatted date from :value with an info icon and :tooltip"
+  [{value :value label :label tooltip :tooltip}]
+  [:div
+   {:style (merge {:color theme-colors/gray} common-styles/body-2-bold)
+    :data-cy "date-label-component"}
+   label
+   [:div
+    {:style common-styles/body-1-bold}
+    [popper-tooltip tooltip
+     [icons/action-info-outlined
```
jkorpelin, line 607: Not best practice to position elements with some arbitrary positional pixel values. 

if it's meant to align the icon with the text you could have the wrapping div have the `common-styles/flex-row-center`

app/frontend/src/cljs/teet/ui/common.cljs
```diff
@@ -592,3 +593,19 @@
...
+      {:style (common-styles/info-icon-style theme-colors/blue-tab "5px")}]]
```
jkorpelin, line 608: If you want to use this instead of the flex-row this `info-icon-style` should get a map as an arg so the "5px" can be a named value so it would be more clear as to what that means in the calling place.

app/frontend/src/cljs/teet/ui/common.cljs
aleksievelizhanin: Just adding common-styles/flex-row-center doesn't work. the exception is thrown by align function

jkorpelin: Yea it requires there to be a wrapping div to which the flex-row is given.

aleksievelizhanin: Fixed by wrapping into Grid container and using alignment center.

app/frontend/src/cljs/teet/ui/common.cljs
```diff
@@ -592,3 +593,20 @@
         [Paper
          (mapc (r/partial context-menu-item toggle!)
                (remove nil? items))]]]]]))
+
+(defn date-label-component
+  "Shows :label with formatted date from :value with an info icon and :tooltip"
+  [{value :value label :label tooltip :tooltip}]
+  [Grid
```
jkorpelin, line 600: Why is this a grid? If you only have 1 grid item the whole grid is unnecessary.

app/frontend/src/cljs/teet/ui/common.cljs
```diff
@@ -592,3 +593,20 @@
...
+   {:container true
+    :data-cy "date-label-component"} label
+   [Grid
+    {:container true
+     :align-items :center}
+    [popper-tooltip tooltip
+     [icons/action-info-outlined
+      {:style {:color theme-colors/blue-tab}}]]
```
jkorpelin, line 608: Icons can be given an option :color :primary to make them use the primary color selected in the theme templates. Is this icon supposed to have some other color than the primary one?

app/frontend/src/cljs/teet/ui/common.cljs
aleksievelizhanin: Grid is used to apply alignment in the current structure of elements. Changing to :div causes errors applying flex alignment.

app/frontend/src/cljs/teet/ui/common.cljs
aleksievelizhanin: This color theme-colors/blue-tab is not the same as :primary This was specific deisgn requirement to have this color applied to the icon.

app/frontend/src/cljs/teet/ui/common.cljs
```diff
@@ -592,3 +593,19 @@
...
+  [{value :value label :label tooltip :tooltip}]
+  [:div
+   {:data-cy "date-label-component"}
+   [:label {:class (<class common-styles/input-label-style false false)}
+    [typography/Text2Bold label]]
+   [:div
+    {:class (<class common-styles/flex-row-center)}
+    [popper-tooltip tooltip
+     [icons/action-info-outlined
+      {:style {:color :prime}}]]
```
jkorpelin, line 608: It needs to be `{:color :primary}`

app/frontend/src/cljs/teet/ui/common.cljs
```diff
@@ -592,3 +593,19 @@
...
+    [:div
+     {:style {:padding-left "0.3em"}}]
```
jkorpelin, line 610: Either this div needs to wrap the text or the padding can be given to the text element it self.

jkorpelin (approved):

PR #421, opened 2021-04-13
Danny does not have admin privileges, so let's try with Benjamin Boss

kristjanZ:


kristjanZ:
@tatut how can we unify that this "CI environment" and the DEV environment have the same mock users with the same privileges?
And no, it does not make sense to give Danny admin privileges in DEV.
image
the privileges should match the names/companies.
Danny - manager
Carla - external consultant
Benjamin - admin
Irma - internal consultant

kristjanZ:
Created https://jira.mkm.ee/browse/TEET-1550 closing this PR.

PR #426, opened 2021-04-15
TEET-1469 ROTL: support referencing other attributes in min/max values

outamaa:
This PR adds fronend validation of min and max values by referencing an attribute of the edited component or any of its parents.


app/backend/src/clj/teet/environment.clj
```diff
@@ -311,18 +311,24 @@
...
-            (migrate-asset-db conn)
-            (reset! asset-db-migrated? true))
-          conn))))
+  (let [db-name (asset-db-name)]
+    (cond (not (feature-enabled? :asset-db))
+          (throw (ex-info "Asset database is not enabled" {:missing-feature-flag :asset-db}))
+
+          (nil? db-name)
+          (throw (ex-info "Asset database name is not defined" {}))
+
```
outamaa, line 320: This was added since it threw me off guard.

app/build-tools/src/teet/asset_schema.clj
```diff
@@ -131,6 +135,49 @@
 (defn- map-by-name [things]
   (into {} (map (juxt :name identity)) things))
 
+(defn- attribute-name
+  "Builds the attribute name from its `:name` and `:ctype`.
+   `(attribute-name {:name :barrier :ctype :barrierworkingwidth}) => :barrier/barrierworkingwidth`"
+  ;; TODO: Why is datatype needed here?
+  [{n :name :keys [ctype datatype] :as _attr}]
+  (when (and n ctype datatype)
+    (keyword (name ctype) (name n))))
```
outamaa, line 144: This was refactored into its own function. Is there a reason for checking for the existence of `datatype`?

tatut: the existence checks are needed as the rotl may be incomplete and you'd get errors about entity only being used as an id

tatut (approved):

PR #432, opened 2021-04-20
TEET-1308 version schema

tatut:
Add BOQ version that can be locked with test that locked version actually prevents changes.

### DIFF
--- a/app/backend/src/clj/teet/asset/asset_commands.clj
+++ b/app/backend/src/clj/teet/asset/asset_commands.clj
@@ -2,16 +2,16 @@
   "Commands to store asset information"
   (:require [teet.db-api.core :refer [defcommand tx] :as db-api]
             [datomic.client.api :as d]
-            [teet.environment :as environment]
-            [clojure.walk :as walk]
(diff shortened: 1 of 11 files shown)


app/backend/src/clj/teet/asset/asset_commands.clj
```diff
@@ -24,7 +24,9 @@
    :authorization {:cost-items/edit-cost-items {}}
    :pre [^{:error :asset-does-not-belong-to-project}
```
outamaa, line 25: Seems `boq-unlocked?` failing implies a different error message.

app/backend/src/clj/teet/db_api/db_api_handlers.clj
```diff
@@ -68,6 +68,9 @@
   (valAt [_ k default-value]
     (deref-delay (get m k default-value))))
 
+(defn ->context-map [m]
+  (ContextMap. m))
+
```
outamaa, line 73: Unused code?

outamaa (approved):
`->context-map` seems unused based on quick grepping. Otherwise LGTM.

PR #433, opened 2021-04-21
Save log file to S3 bucket upon restore completion

aleksievelizhanin:


app/backend/src/clj/teet/backup/backup_ion.clj
```diff
@@ -364,6 +382,7 @@
              prepare-asset-db-for-restore
              restore-tx-file
              migrate
+             log-restore-result
```
tatut, line 385: what about exception case? I think failure should be logged as well.. so anyone polling for this log file knows that the thing failed

app/backend/src/clj/teet/backup/backup_ion.clj
```diff
@@ -342,6 +342,24 @@
...
+(defn log-restore-result
+  "Write log to bucket into the same file name from which the restore was started"
+  [{{bucket :bucket file-key :file-key} :s3}]
+  (try
+    (let [file (java.io.File/createTempFile file-key ".log")]
+      (log/info "Generating restore log file: " (.getAbsolutePath file))
+      (with-open [w (io/writer file :append true)]
+                 (.write w (str "Backup " file-key " was restored successfully.")))
+      (with-open [in (io/input-stream file)]
+                 (s3/write-file-to-s3
```
tatut, line 354: indent body of `with-` macros at 2

tatut (changes requested):
log failure

app/backend/src/clj/teet/backup/backup_ion.clj
aleksievelizhanin: - Exception in saving file can not be logged as something already wrong in saving to S3 bucket.
- The exception in restore thowns Exception. In this case the log-restore-result function will not be called at all. 
- Is there any other case then failure in restore saved in context, for instance, without throwing an exception?

tatut (approved):

PR #438, opened 2021-04-26
TEET-1511 Update project name in Vektor.io once it is updated in TEET

solita-erkkiv:


app/backend/src/clj/teet/project/project_commands.clj
```diff
@@ -28,14 +30,23 @@
               :thk.project/custom-end-m]
           project-eid))
 
+(defn update-vektorio-project-name? [db project-id project-name]
```
jkorpelin, line 33: The function ending in ? implies that this returns a boolean and doesn't do any side-effects.

app/backend/src/clj/teet/project/project_commands.clj
```diff
@@ -28,14 +30,23 @@
...
+
 (defcommand :thk.project/update
   {:doc "Edit project basic info"
    :context {:keys [conn db user]}
    :payload {id :thk.project/id :as project-form}
    :project-id [:thk.project/id id]
    :authorization {:project/update-info {:eid [:thk.project/id id]
                                          :link :thk.project/owner}}}
-  (let [{db-before :db-before
+  (if (some? (update-vektorio-project-name? db [:thk.project/id id] (:thk.project/project-name project-form)))
```
jkorpelin, line 48: So does this whole `thk.project/update` fail when the vektorio feature folag is not enabled?

erno (approved):
LGTM but let's file a bug for fixing the http status prob solved by :vektorio-status for the other methods too.

app/backend/src/clj/teet/project/project_commands.clj
```diff
@@ -28,14 +30,22 @@
               :thk.project/custom-end-m]
           project-eid))
 
+(defn update-vektorio-project-name? [db project-id project-name]
+  (let [vektorio-enabled? (environment/feature-enabled? :vektorio)
+        vektorio-config (environment/config-value :vektorio)]
+      (if vektorio-enabled?
+            (vektorio/update-project-in-vektorio! db vektorio-config project-id project-name)
+            true)))
```
jkorpelin, line 38: Now you are asking `update-vektorio-project-name?` and returning a `true` when the vektorio is not enabled?
You should probably create a `maybe-update-vektorio-project-name` function and call it after the transaction for the project updating has happened.

PR #442, opened 2021-04-26
TEET-1272 Update model metadata on vektor.io if file metadata is upda…

solita-erkkiv:
…ted in TEET


app/backend/src/clj/teet/file/file_commands.clj
```diff
@@ -277,9 +277,16 @@
                  :file-seen/file+user [file-id user-id]
                  :file-seen/seen-at (java.util.Date.)})]})
 
+(defn maybe-update-vektorio-model-name? [conn db id file]
```
erno, line 280:
```suggestion
(defn maybe-update-vektorio-model-name! [conn db id file]
```

PR #443, opened 2021-04-26
Teet 1132 standalone deploy script

erno:


app/backend/deps.edn
```diff
@@ -1,6 +1,7 @@
 {:deps {http-kit {:mvn/version "2.5.3"}
         ;; Need clj-http for vektorio file uploads
         clj-http/clj-http {:mvn/version "3.12.0"}
+        ring/ring-jetty-adapter {:mvn/version "1.9.3"}
```
tatut, line 4: should this be only for dev dependencies?

app/backend/src/clj/teet/main.clj
```diff
@@ -40,14 +45,24 @@
               (login-fake-routes/fake-login-routes)))
           (db-api-dev/db-api-routes)
           (routes/teet-routes config))
+         wrap-nil
```
tatut, line 48: why is this needed? isn't 404 automatic if there is no response by any handler

ci/scripts/install-standalone-vm.bash
```diff
@@ -0,0 +1,441 @@
...
+# script that installs teet app, postgresql db, datomic local, postgrest api on one vm.
+#
+# usage:
+# run this from shell on your dev machine:
+# ci/scripts/install-standalone-vm.bash launchvm <ssh-key-name>
+# where ssh key name a key name from the registered ec2 key pairs in the aws env.
+# This will start the vm, detect what git branch you are on and try to check out that same branch at the remote vm.
+
+
+MACHINE_ID=b # pick one from a-e that doesn't conflict with someone else
```
tatut, line 12: how should this be picked?

ci/scripts/install-standalone-vm.bash
```diff
@@ -0,0 +1,441 @@
...
+                     :data-frisk
+                     :my-role-display
+                     :dummy-login
+                     :cooperation
+                     :admin-inspector
+                     ; :vektorio
+                     :asset-db
+                     :cost-items
+                     :land-owner-opinions
+                     :assetmanager}
```
tatut, line 83: how do we remember to update the list of enabled features when we add new ones?

app/backend/src/clj/teet/main.clj
erno: Jetty gives you a NPE in this case.

ci/scripts/install-standalone-vm.bash
erno: Good question, I thought about a few different ways to solve it but didn't come up with a good idea yet.

ci/scripts/install-standalone-vm.bash
erno: ah, I thought I was passing this in as an env var, will fix.

app/backend/src/clj/teet/main.clj
```diff
@@ -1,5 +1,5 @@
 (ns teet.main
```
erno, line 1: show version banner with git hash / git_branch

ci/scripts/install-standalone-vm.bash
```diff
@@ -0,0 +1,442 @@
...
+	exit 1
+    fi
+    shift
+    # assume we're running in codebuild or dev workstation
+    run-in-ec2 "$@" # keypair name passed as arg
+    
+else
+    # we're running as the cloud-init script (aka user-data script) inside the vm
+    # - output will appear as cloud-init messages in /var/log/syslog
+    echo starting teet app install
```
erno, line 439: add slack message that gives the url for the launched vm

erno: ran out of time to do the slack message, let's do it later.

PR #448, opened 2021-04-27
TEET-1580 grouping and filtering totals

tatut:


app/common/src/cljc/teet/util/collection.cljc
```diff
@@ -283,3 +283,30 @@
...
+              (into path [here c])
+
+              ;; not found here, recurse
+              (first
+               (for [sub cs
+                     :let [sub-path (containing (conj path here) sub)]
+                     :when sub-path]
+                 sub-path)))))]
+
+    (containing [] form)))
```
outamaa, line 312: Good, highly reusable.

app/frontend/src/cljs/teet/asset/cost_items_controller.cljs
```diff
@@ -470,3 +478,24 @@
...
     (t/fx app
           common-controller/refresh-fx)))
+
+(defn filtered-cost-group-totals
+  "Return ATL item that is being filtered by and cost group totals that match
+  the filter.
+
+  Adds :ui/group to each item that contains a vector of type hierarchy."
+  [app atl cost-group-totals]
+  (let [kw (some-> app (get-in [:query :filter]) cljs.reader/read-string)
```
outamaa, line 488: Had to check that this can't be used to read arbitrary Clojurescript code. :D It's essentially an edn reader.

tatut: I had the same concern and checked, perhaps should comment that so future readers don't get raised blood pressure

PR #454, opened 2021-04-29
TEET-1584 rotl viewer 2 panels

tatut:
Make ROTL viewer 2 panels: left side navigates the hierarchy and right side shows details of clicked item.

### DIFF
--- a/app/frontend/src/cljs/teet/asset/asset_library_view.cljs
+++ b/app/frontend/src/cljs/teet/asset/asset_library_view.cljs
@@ -5,12 +5,17 @@
             [teet.localization :as localization :refer [tr]]
             [teet.ui.table :as table]
             [teet.util.collection :as cu]
-            [teet.ui.material-ui :refer [Card CardHeader CardContent IconButton]]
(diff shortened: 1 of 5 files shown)


app/frontend/src/cljs/teet/asset/asset_library_view.cljs
```diff
@@ -85,47 +103,106 @@
...
+    [child-links (:fclass/_fgroup fgroup)]]])
+
+(defn- rotl-type-badge-style []
+  {:font-size "70%"
+   :font-variant :all-small-caps
+   :margin-left "1rem"
+   :padding "0px 5px 0px 5px"
+   :border-radius "10px"
+   :display :inline-block
+   :background theme-colors/gray-lightest})
```
outamaa, line 132: Wonder if the text style defined here was one of those defined in common-styles.

outamaa (approved):

app/frontend/src/cljs/teet/asset/asset_library_view.cljs
tatut: I couldn't find it, we have some badge styles in notification view but it's different.

PR #457, opened 2021-04-30
Teet 1291 remove activity procurement no

aleksievelizhanin:


app/backend/src/clj/teet/migration/activity_procurement_info.clj
```diff
@@ -21,3 +21,16 @@
...
+(defn retract-procurement-info-from-activity [conn]
+  (let [activities (d/q '[:find ?a
+                          :where [?a :activity/procurement-id _]
+                          [?a :activity/procurement-nr _]]
+                     (d/db conn))]
+    (d/transact conn
+      {:tx-data (for [[activity-id] activities]
+                  [:db/retract activity-id :activity/procurement-id])})
+
+    (d/transact conn
```
jkorpelin, line 34: why is this two separate transactions instead of just one? And shouldn't the retract be given the attribute value that is being retracted as well https://docs.datomic.com/cloud/tutorial/retract.html ?

app/backend/test/teet/activity_commands_test.clj
```diff
@@ -197,30 +197,12 @@
...
-    (tu/tx {:db/id (tu/get-data :new-activity-id)
-            :activity/procurement-nr "123 R 456"})
-
-    (is (thrown? Exception
-                 (tu/local-command tu/mock-user-boss
-                                   :activity/delete
-                                   {:db/id (tu/get-data :new-activity-id)})))
-
-    (is (not (:meta/deleted? (du/entity (tu/db)
-                                        (tu/get-data :new-activity-id)))))
```
jkorpelin, line 214: you could add a check in this checks place that makes sure non admins can't delete activities.

aleksievelizhanin: So, it should be added to AC first

jkorpelin: > Admins can delete all activities that don't have files (remove the procurement constraint mentioned above)

Is it not in the ac?

app/backend/src/clj/teet/migration/activity_procurement_info.clj
aleksievelizhanin: Two transactions implemented to avoid mergeing two lists for one transaction. Could probably be re-factored into one.
My understanding of this tutorial that we can add attribute value to be retracted, only if we care about other attributes values, which we want to keep. In this case we want to retract all values of procurement-id, and procurement-nr, don't we?

app/backend/test/teet/activity_commands_test.clj
aleksievelizhanin: It is in the AC. As we agreed we don't test auth. matrix?

app/backend/src/clj/teet/migration/activity_procurement_info.clj
```diff
@@ -21,3 +21,15 @@
...
+
+(defn retract-procurement-info-from-activity [conn]
+  (let [activities (d/q '[:find ?a
+                          :where [?a :activity/procurement-id _]
+                          [?a :activity/procurement-nr _]]
+                     (d/db conn))
+        procurement-ids (for [[activity-id] activities]
+                          [:db/retract activity-id :activity/procurement-id])
+        procurement-nrs (for [[activity-id] activities]
+          [:db/retract activity-id :activity/procurement-nr])]
```
jkorpelin, line 33: `(let [activities (->> (d/q '[:find ?a
                              :where [?a :activity/procurement-id _]
                              [?a :activity/procurement-nr _]]
                            (d/db conn))
                       (mapv first))
        procurement-ids (for [activity-id activities]
                          [:db/retract activity-id :activity/procurement-id])
        procurement-nrs (for [activity-id activities]
                          [:db/retract activity-id :activity/procurement-nr])]`

I think it's a bit more clear if you mapv first out of the activities list and then you don't need the destructuring of the list elements in the for loop.

jkorpelin (approved):
Other than the one comment looks good to me

app/backend/src/clj/teet/thk/thk_mapping.clj
```diff
@@ -150,9 +150,7 @@
    "activity_thkupdstamp"
    "activity_teetupdstamp"
    "activity_teetdelstamp"
-   "activity_cost"
-   "activity_procurementno"
```
kristjanZ, line 154: does this mean that these columns are removed from the teet-thk export csv?

aleksievelizhanin: not necessary, it only means we are not using information from them

PR #469, opened 2021-05-05
Teet 1291 fix activity edit modal

aleksievelizhanin:


ci/browser-tests/cypress/integration/activity_spec.js
```diff
@@ -0,0 +1,17 @@
...
+    before(() => {
+        cy.dummyLogin("benjamin")
+        cy.selectLanguage("#EN")
+        cy.projectByName("integration test project")
+        cy.get("li a").contains("Detailed design").click()
+    })
+
+    context("Edit Activity", function() {
+        it("can be edited", function() {
+            cy.get("button[class=\"MuiButtonBase-root MuiButton-root " +
```
kristjanZ, line 11: This "calss" approach doesn't seem to be better than going with The "Edit" text?

jkorpelin: Should insert an id or a data attribute specifically for this case.

PR #479, opened 2021-05-07
TEET-1479 Handle duplicate files in export zip

outamaa:
Activity export was fixed earlier. This PR adds duplicate file name renaming to task export. This is needed because earlier, duplicate filenames in one task were allowed, and such tasks exist in production.

### DIFF
--- a/app/backend/src/clj/teet/file/file_export.clj
+++ b/app/backend/src/clj/teet/file/file_export.clj
@@ -12,7 +12,7 @@
             [taoensso.timbre :as log])
   (:import (java.util.zip ZipOutputStream ZipEntry)))
 
-(defn task-zip-entries
+(defn- task-zip-entries
   "Return zip file entries for current versions of all files in the given task."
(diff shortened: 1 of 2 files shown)


app/backend/src/clj/teet/file/file_export.clj
```diff
@@ -57,6 +57,31 @@
...
            (log/info e "io/copy failed in zip creation")))))))
 
+(defn- file-name-with-sequence-number [seq fn]
+  (let [{:keys [extension description]}
+        (filename-metadata/name->description-and-extension fn)
+        new-description (str description " (" seq ")")]
+
+    (if (not-empty extension)
+      (str new-description "." extension)
+      ;; else:
```
aleksievelizhanin, line 67: I would refactor this ext check into the function with descriptive name to avoid ;; ;; comments

aleksievelizhanin (approved):
Looks legit. As a minor improvement I would do less comments and more functions with descriptive names.

PR #491, opened 2021-05-11
Teet 1237 contracts in project detail

jkorpelin:


app/backend/src/clj/teet/contract/contracts_queries.clj
```diff
@@ -112,3 +113,14 @@
...
+
+(defquery :contracts/project-related-contracts
+  {:doc "Return a list of contracts related to the given project"
+   :context {db :db}
+   :args {project-id :thk.project/id}
+   :project-id project-id
+   :authorization {}}
+  (->> (contract-db/project-related-contracts db [:thk.project/id project-id])
+       (sort-by
+         #((:thk.contract/status %) contract-model/contract-status-order))
```
tatut, line 125: would be clearer imo to `(sort-by (comp contract-model/contract-status-order :thk.contract/status) ...)`

And that won't crash if there is no status.

tatut (commented):
one tiny comment, looks good

app/backend/src/clj/teet/contract/contracts_queries.clj
jkorpelin: Good catch a lot clearer idd

PR #507, opened 2021-05-17
TEET-1577 changes zip download url generation to make sure it stays v…

jkorpelin:
…alid

### DIFF
--- a/app/backend/src/clj/teet/file/file_queries.clj
+++ b/app/backend/src/clj/teet/file/file_queries.clj
@@ -113,7 +113,7 @@
       ;; know that filename is not valid
       {})))
 
-(defn valid-filename?
+(defn- valid-filename?
   [filename]
   (let [{extension :extension
          description :description} (filename-metadata/name->description-and-extension filename)]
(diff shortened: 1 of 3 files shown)


app/backend/src/clj/teet/file/file_queries.clj
```diff
@@ -114,3 +112,28 @@
       ;; If metadata can't be parsed, return empty map, frontend will
       ;; know that filename is not valid
       {})))
+
+(defn valid-filename?
```
tatut, line 116: valid filename for what? if this is only used for the exported zip, the name could reflect that.
It could also be more strict, so that the filename part must be valid uuid.
should also be private

tatut (commented):
minor comment, otherwise looks legit

app/backend/src/clj/teet/file/file_queries.clj
jkorpelin: Added it as a precaution since the endpoint is open and the filename goes straight to the content-disposition, so it can't be messed around with. Maybe the s3-key should have some checking as well?

erno: Could also check for metacharacters (newline at least) that could be used to do header injection like this: https://owasp.org/www-community/attacks/HTTP_Response_Splitting

jkorpelin: The valid-chars-in-description? should not allow any extra characters in the filename

PR #509, opened 2021-05-18
TEET-1625 Cost item totals: price edit bug "server error"

outamaa:
This PR implements the following fixes and improvements:
- Handle empty input by removing the price information
- Properly format field value after successful update
- Disable negative values for prices
- Set scale of prices to 2 decimal places, as Datomic strongly recommends having the same scale for all values of a given attribute.
- Fix case where price couldn't be updated back to previous value
- Fix case where text field remained disabled after attempting to input invalid price
- Fix case where text field remained disabled after successful update
- Fix case where inputting more than 2 decimal numbers results in updates failing due to failed CAS

### DIFF
--- a/app/frontend/src/cljs/teet/asset/cost_items_controller.cljs
+++ b/app/frontend/src/cljs/teet/asset/cost_items_controller.cljs
@@ -99,8 +99,9 @@
 
 (defrecord AddComponent [type])
 
-(defrecord SaveCostGroupPrice [cost-group price])
-(defrecord SaveCostGroupPriceResponse [response])
+(defrecord SaveCostGroupPrice [finish-saving! cost-group price])
+(defrecord SaveCostGroupPriceResponse [finish-saving! response])
+(defrecord SaveCostGroupPriceError [finish-saving! error])
 
 ;; Locking and versioning
 (defrecord SaveBOQVersion [callback form-data])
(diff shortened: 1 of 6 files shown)


app/frontend/src/cljs/teet/asset/cost_items_view.cljs
```diff
@@ -905,30 +906,34 @@
...
                          (reset! saving? true)
-                         (e! (cost-items-controller/->SaveCostGroupPrice row @price))))
-               save-on-enter! (fn [row e]
+                         (e! (cost-items-controller/->SaveCostGroupPrice finish-saving! row @price))))
+               save-on-enter! (fn [current-value row e]
                                 (when (= "Enter" (.-key e))
-                                  (save! row)))]
-    (when (not= value initial-value)
-      (reset! saving? false))
+                                  (save! current-value row)))]
```
outamaa, line 925: In the above, I tried to keep the fix as local as possible. Don't know if this was a good decision.

app/frontend/src/cljs/teet/asset/cost_items_controller.cljs
```diff
@@ -518,15 +519,24 @@
                                          :total-cost
                                          :quantity-unit :type
                                          :ui/group)
-                     :price price}
-           :result-event ->SaveCostGroupPriceResponse}))
+                     ;; Interpret empty string as 0
+                     :price (if (str/blank? price) "0" price)}
```
tatut, line 523: I don't understand why empty is zero?

There's a difference between something being free and the price not being known for it.
You should be able to retract a price without setting it to zero.

outamaa: Talked about this in Slack, let's change to retract.

PR #510, opened 2021-05-18
Add db name to connection to be able to connect to other than current…

aleksievelizhanin:
… db (for Tableau, for example)

### DIFF
--- a/app/backend/src/clj/teet/environment.clj
+++ b/app/backend/src/clj/teet/environment.clj
@@ -299,17 +299,19 @@
 
 (defn datomic-connection
   "Returns thread bound connection or creates a new one."
-  []
-  (or *connection*
-      (let [db (db-name)
-            client (datomic-client)
-            db-status (ensure-database client db)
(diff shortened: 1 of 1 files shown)


app/backend/src/clj/teet/environment.clj
```diff
@@ -299,17 +299,19 @@
...
-            conn (d/connect client {:db-name db})]
-        (log/info "Using database: " db db-status)
-        (when-not @db-migrated?
-          (migrate conn @schema)
-          (reset! db-migrated? true))
-        conn)))
+  ([]
+   (datomic-connection [(db-name)]))
+  ([database-name]
+   (or *connection*
```
tatut, line 305: what if thread bound `*connection*` is to a database that has a different name?

aleksievelizhanin: I can't imagine the situation. Reasonbly to assume that the caller of this function with database-name param must have checked the name validity..

PR #525, opened 2021-05-21
Direct JDBC calls instead of PostgREST

tatut:
This is a Work In Progress... do not merge, it needs infrastructure changes to enable the direct connection from compute group nodes.

For discussion an planning.

- New way to get pg connection from environment
- Add pg connection to query/command context with automatic cleanup
- Add geo queries for GeoJSON and MVT endpoints
- Use direct calls to update project geometries

To decide: 
- datasource import also uses the API, do we want to change that to direct JDBC as well?

### DIFF
--- a/app/backend/src/clj/teet/asset/asset_road_registry_import.clj
+++ b/app/backend/src/clj/teet/asset/asset_road_registry_import.clj
@@ -1,0 +1,32 @@
+(ns teet.asset.asset-road-registry-import
+  "Import assets to TEET from road registry WFS.
+  Requires a mapping for an asset type that describes
+  what components are created and what fields go to
+  which attributes."
+  (:require [teet.road.road-query :as road-query]
+            [teet.map.map-services :as map-services]
+            [hiccup.core :as h]))
+
+
+(defn- road-registry-objects [config type gml-area]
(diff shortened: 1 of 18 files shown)


app/backend/src/clj/teet/asset/asset_road_registry_import.clj
```diff
@@ -0,0 +1,32 @@
+(ns teet.asset.asset-road-registry-import
```
tatut, line 1: Disregard, this will be merged before this in another branch... mistakenly based this branch on a WIP

app/backend/src/clj/teet/db_api/db_api_handlers.clj
```diff
@@ -75,51 +76,64 @@
...
+                   pg (delay
+                        (let [c (environment/get-pg-connection)]
+                          (swap! cleanup conj
+                                 #(do
+                                    (log/debug "Closing PostgreSQL connection" c)
+                                    (.close c)))
+                          {:connection c}))
+                   ctx {:conn conn
+                        :db db
+                        :pg pg
```
tatut, line 97: we add a `:pg` key to context, which if used will get a connection from the pool and register cleanup function that is run below in the finally block.

PR #529, opened 2021-05-25
Teet 1671 import owner supervision and audit tasks

aleksievelizhanin:
Add new tasks from THK activities


app/backend/src/clj/teet/thk/thk_import.clj
```diff
@@ -180,8 +181,14 @@
                            name :activity/name
                            task-id :activity/task-id
                            :as activity} activities
-                          ;; Only process activities here
-                          :when (and id name (nil? task-id))
+                          ;; Only process activities here except 2 with activity_typefk 4006 and 4009
```
erno, line 184: Could expand comment, why not those activity types?

app/backend/src/clj/teet/thk/thk_import.clj
```diff
@@ -222,6 +229,111 @@
...
+               :in $ ?e] db activity-eid)]
+    (when
+      (some? query-result)
+      (:activity/tasks (ffirst query-result)))))
+
+(defn- get-existing-task-db-id
+  "Search for the existing task with the same type and start and end dates"
+  [db construction-activity-id task-type start-date end-date]
+  (first (reduce
+           #(if
```
erno, line 276: Could be more readable with fn that had named args. also (identity x) is same as x.

app/backend/src/clj/teet/thk/thk_import.clj
aleksievelizhanin: it is a requirement of TEET-1671
Import owners supervision and audit tasks and contracts from thk-teet csv to teet

app/backend/src/clj/teet/thk/thk_import.clj
```diff
@@ -224,6 +232,112 @@
...
+        (d/q '[:find (pull ?e [:activity/tasks])
+               :in $ ?e] db activity-eid)]
+    (when
+      (some? query-result)
+      (:activity/tasks (ffirst query-result)))))
+
+(defn- get-existing-task-db-id
+  "Search for the existing task with the same type and start and end dates"
+  [db construction-activity-id task-type start-date end-date]
+  (first (reduce
```
tatut, line 278: It looks like you could replace get-task-type get-activity-task-eids and this query with one `d/q` this seems difficult to read for me

app/backend/src/clj/teet/thk/thk_import.clj
```diff
@@ -224,6 +232,112 @@
...
+      :activity.name/road-safety-audit :task.type/road-safety-audit
+      (log/info "new task type unknown for activity " activity-name))))
+
+(defn add-task-from-thk
+  "Returns TX data for new task with given params for given construction activity-id"
+  [db thk-activity-task-data construction-activity-id ]
+  (let [start-date (:activity/estimated-start-date thk-activity-task-data)
+        end-date (:activity/estimated-end-date thk-activity-task-data)
+        thk-activity-status (:activity/status thk-activity-task-data)
+        task-type (get-new-task-type thk-activity-task-data)
```
tatut, line 303: what happens if `task-type` is `nil`?  (the last case in `get-new-task-type`

app/backend/src/clj/teet/thk/thk_import.clj
```diff
@@ -224,6 +232,112 @@
...
+        existing-task-eid (get-existing-task-db-id db construction-activity-id task-type
+                               start-date end-date)
+        id-placeholder (if (some? existing-task-eid)
+                         existing-task-eid
+                         (str "NEW-TASK-" (name :task.group/construction) "-" task-type
+                           "-" (integration-id/unused-random-small-uuid db)))
+        task-tx-data (merge
+                       {:db/id id-placeholder
+                        :task/group (if (= :task.type/road-safety-audit task-type)
```
tatut, line 312: should this be a case to enumerate all possibilities. Now if type is not road safety audit, it will be the other one. 
seems easy to miss if there are ever new types that need to be supported

tatut (commented):
Some details that could be clearer, but looks legit

app/backend/src/clj/teet/thk/thk_import.clj
aleksievelizhanin: Replaced!

app/backend/src/clj/teet/thk/thk_import.clj
aleksievelizhanin: Datomic tx will fail. Changed it to exception throwing with some activity info in details.

app/backend/src/clj/teet/thk/thk_import.clj
aleksievelizhanin: Refactored into the case.

app/backend/src/clj/teet/thk/thk_mapping.clj
```diff
@@ -68,6 +68,8 @@
    "4003" :activity.name/detailed-design ;; Pöhiprojekt
    "4004" :activity.name/land-acquisition ;; Maaost
    "4005" :activity.name/construction ;; Teostus
+   "4006" :activity.name/owners-supervision ;; OJV
+   "4009" :activity.name/road-safety-audit ;; Tee Ohutus Audit
```
jkorpelin, line 72: Should these be mapped straight to task.types since they are not activities in TEET?

aleksievelizhanin: They are activities in .csv and this mapping for parsing csv, so no they should not be mapped to task.types as they are not task types in .csv

jkorpelin (approved):

PR #563, opened 2021-06-01
Cost item totals performance fix

tatut:


app/backend/src/clj/teet/asset/asset_db.clj
```diff
@@ -342,7 +338,7 @@
   (let [items (cost-group-attrs-q
                db (if (some? oids)
                     oids
-                    (project-asset-and-component-oids db thk-project-id)))
+                    (project-assets-and-components db thk-project-id)))
```
outamaa, line 341: I liked the `-oids` suffix.

outamaa (approved):

app/backend/src/clj/teet/asset/asset_db.clj
tatut: yeah, but it was the odd one... the others also returned OIDs but didn't have it in the name. this is now more consistent.

PR #571, opened 2021-06-02
TEET-1495 : add activity managers to contract list query and display them to contract card

jalef:


app/backend/src/clj/teet/contract/contract_db.clj
```diff
@@ -119,6 +119,13 @@
   [[contract status]]
   (assoc contract :thk.contract/status status))
 
+(defn contract-with-status-and-manager
+  "Used after datomic query which returns [contract contract-status manager]"
+  [[contract status manager]]
```
jkorpelin, line 124: the manager is not needed here anymore since the manager is fetched in the pull pattern of the contract.

jkorpelin (changes requested):
The contract-with-status-and-manager is not necessary since the value doesn't come as separate find value from the query, but is instead a part of the contract pull query.

jkorpelin (approved):

PR #574, opened 2021-06-03
TEET-1652 BIM redirect link vs JWT expiration

erno:


erno:
Will make a new PR withut the standalone vm script merge

PR #576, opened 2021-06-03
Check Procurement No starts from digit to show the link

aleksievelizhanin:
Add condition to show link to Procurement state registry - only if Procurement No starts with a digit the link will be generated, overwise just a :span with Procurement No. 

Note! may require some addtional styling then there is no link.


app/frontend/src/cljs/teet/contract/contract_common.cljs
```diff
@@ -12,12 +12,16 @@
...
             [teet.ui.format :as format]
             [teet.util.euro :as euro]
+            [clojure.edn :as edn]
             [teet.user.user-model :as user-model]))
 
 (defn contract-procurement-link
   [{:thk.contract/keys [procurement-number]}]
-  [common/external-contract-link {:href (str (environment/config-value :contract :state-procurement-url) procurement-number)}
-   (str (tr [:contracts :state-procurement-link]) " " procurement-number)])
+  (if (js/Number.isInteger (edn/read-string (subs procurement-number 0 1)))
```
jkorpelin, line 20: Is the edn dependency necessary? Would just `js/parseInt` and then cljs core `integer?` be enough?

jkorpelin (approved):
Other than the minor comment lgtm

jkorpelin:
Should the link just be not shown if the id is wrong?

aleksievelizhanin:
The AC requiers to show the Procurement No (not as a link) if it is started not from digit, not just hide the link..

PR #577, opened 2021-06-03
Teet 1073 - Task part review fixes

solita-erkkiv:


app/frontend/src/cljs/teet/task/task_view.cljs
```diff
@@ -272,7 +272,7 @@
               :button-text (tr [:task-part :reject])}]]]
 
           :file.part.status/completed
-          [when-authorized :task/reopen-task task
+          [when-authorized :task/task-part-reopen task
```
aleksievelizhanin, line 275: old variant sounds grammatically better - authorized to do something. If the only change "task" -> "task-part" then [when-authorized :task/reopen-task-part 
IMO

PR #592, opened 2021-06-08
Increase Cypress login timeout to prevent flaky tests

outamaa:
This PR will increase the Cypress login timeout to 30 seconds.


aleksievelizhanin (commented):
From default 4 sec we are going to wait 30 sec for login. Looks like we are working around the bigger problem (slow server). Looks legit, otherwise.

aleksievelizhanin (approved):

PR #595, opened 2021-06-08
Update existing contracts

aleksievelizhanin:


app/backend/src/clj/teet/thk/thk_import.clj
```diff
@@ -354,32 +354,34 @@
...
+                          contract-db-id)
+                 :thk.contract/targets targets}
+           region-tx
+           (select-keys contract-info [:thk.contract/procurement-id
+                                       :thk.contract/name
+                                       :thk.contract/part-name
+                                       :thk.contract/procurement-number
+                                       :thk.contract/procurement-part-id
+                                       :thk.contract/type])
+           (meta-model/system-created))]
```
jkorpelin, line 383: Should you add system-modified timestamp if there is an existing db/id?

aleksievelizhanin: good point

jkorpelin, line 383: Only merge this in, if there is no contract-db-id

tatut (approved):

PR #609, opened 2021-06-10
TEET-1671 Import owners supervision and audit tasks and contracts from thk-teet csv to teet

aleksievelizhanin:
- Tasks actual start and end date have been imported
- thk.activity/id
- create-at done only once in the first time when task is created


tatut (approved):
lgtm

PR #620, opened 2021-06-14
Fix/contract UI fixes

jalef:


app/frontend/src/cljs/teet/contract/contract_common.cljs
```diff
@@ -12,18 +12,23 @@
...
+      [common/external-contract-link
+       {:href (str (environment/config-value :contract :state-procurement-url) procurement-number "/general-info")}
+       procurement-display-text]
+      [typography/SmallText {:style {:display :flex
+                                     :align-items :center
+                                     :margin-right "1rem"}}
+       procurement-display-text])))
 
 (defn contract-external-link
+  [{:thk.contract/keys [external-link procurement-number]}]
```
jkorpelin, line 31: this seems like a typo? or accidental copy of the parameter row.

jalef: totally missed that. Fixed now

PR #622, opened 2021-06-14
Edit company 

aleksievelizhanin:


app/backend/resources/schema.edn
```diff
@@ -2123,5 +2123,13 @@
 
  {:db/ident :migration/subcontractor-role
   :txes [[{:db/doc "Not selectable in UI. Added automatically to subcontractor employees"
-           :db/ident :company-contract-employee.role/subcontractor}]]}]
+           :db/ident :company-contract-employee.role/subcontractor}]]}
+
+ {:db/ident :migration/change-cardinality-many-to-one-company-emails
+  :txes [[{:db/ident :company/emails :db/cardinality :db.cardinality/many}
```
jkorpelin, line 2129: Is it not enough to just transact the `:db.cardinality/one`

aleksievelizhanin: based on https://docs.datomic.com/cloud/schema/schema-change.html there 2 things to be done
- "When changing from a multi-valued attribute to a single-valued attribute, there must be at most a single value for every entity in the set of current assertions"
- transact the change of cardinality

I didn't include check for values as they are not a part of the schema

jkorpelin: >  there must be at most a single value for every entity in the set of current assertions; otherwise, the change will not be accepted and the transaction will fail.

The transaction will fail for dev. Maybe we should have a migration that just retracts the value first as there is nothing in the dev that would need to be saved?

jkorpelin: > there must be at most a single value for every entity in the set of current assertions; otherwise, the change will not be accepted and the transaction will fail.

So I'm assuming the migration fails in dev? Maybe we should retract all the emails and phones in another migration that's run first?

app/backend/src/clj/teet/contract/contract_commands.clj
```diff
@@ -32,8 +33,10 @@
          (company-db/business-registry-code-unique?
            db
            (company-model/company-business-registry-id-with-country-code form-data))]}
-  (let [company-fields (-> (select-keys form-data [:company/country :company/emails :company/phone-numbers
+  (let [company-fields (-> (select-keys form-data [:company/country
                                                    :company/business-registry-code :company/name])
+                           (assoc :company/emails (str/join "" (:company/emails form-data)))
+                           (assoc :company/phone-numbers (str/join "" (:company/phone-numbers form-data)))
```
jkorpelin, line 39: should these be joined by ", " instead of just "" ? Also `(update :company/emails #(str/join ", " %))` works a bit simpler.

app/backend/src/clj/teet/contract/contract_commands.clj
```diff
@@ -61,6 +64,43 @@
      (merge tempids
             {:company-contract-id new-company-contract-id})))
 
+(defcommand :thk.contract/save-contract-partner-company
```
jkorpelin, line 67: Should this be called `edit-contract-partner-company` if it's used for editing?

app/backend/src/clj/teet/contract/contract_commands.clj
```diff
@@ -61,6 +64,43 @@
...
+        (:tempids (tx [(list 'teet.contract.contract-tx/update-contract-partner
+                         contract-eid
+                         lead-partner?
+                         company-id
+                         [(merge
+                            company-fields
+                            (meta-model/modification-meta user))
+                          (merge
+                            {:teet/id company-contract-uuid
+                             :company-contract/lead-partner? (if (true? lead-partner?) true false)}
```
jkorpelin, line 100: We should retract the lead-partner? value all together so it's a bit easier to query for it with the `missing?` function similar to the `:meta/deleted?`

app/frontend/src/cljs/teet/contract/contract_partners_view.cljs
```diff
@@ -175,6 +237,73 @@
...
+  (str (tr [:common :last-modified]) " "
+    (if (nil? (:meta/modified-at selected-company))
+      (str (format/date-time (:meta/created-at selected-company)) " "
+        (get-in selected-company [:company-contract/company :meta/creator :user/given-name])
+        " "
+        (get-in selected-company [:company-contract/company :meta/creator :user/family-name]))
+      (str (format/date-time (:meta/modified-at selected-company)) " "
+        (get-in selected-company [:company-contract/company :meta/modifier :user/given-name])
+        " "
+        (get-in selected-company [:company-contract/company :meta/modifier :user/family-name])))))
```
jkorpelin, line 250: We have in the user-model ns a function called user-name, which could be used here.

app/backend/resources/schema.edn
aleksievelizhanin: New singular attributes :company/email and :company/phone-number were added and used instead of plural ones. Thas resolves the migration issue.

app/backend/src/clj/teet/contract/contract_commands.clj
jkorpelin: and similar to when we add a new lead partner.

app/backend/src/clj/teet/contract/contract_commands.clj
aleksievelizhanin: fixed by parsing vector of emails and vector of phone-numbers right after x-road query. And then read/wride as a single string

app/backend/resources/schema.edn
aleksievelizhanin: fixed by using single email and phone-number

app/backend/src/clj/teet/contract/contract_commands.clj
```diff
@@ -63,6 +65,44 @@
...
+   :payload {form-data :form-data
+             contract :contract}
+   :context {:keys [user db]}
+   :project-id nil
+   :authorization {:contracts/contract-editing {}}
+   :pre [^{:error :business-registry-code-in-use}
+         (company-db/business-registry-code-unique?
+           db
+           (company-model/company-business-registry-id-with-country-code form-data))]}
+  (let [company-fields (-> (select-keys
```
jkorpelin, line 79: the arrow seems unnecessary.

app/backend/src/clj/teet/contract/contract_db.clj
```diff
@@ -122,7 +122,20 @@
...
+                         [* {:company-contract/company
+                             [:company/business-registry-code
+                              :company/name
+                              :teet/id
+                              :db/id
+                              :company/phone-number
+                              :company/email
+                              :company/country
+                              :meta/created-at
+                              :meta/modified-at
```
jkorpelin, line 136: the company keys could come from some company-model ns.

app/frontend/src/cljs/teet/common/common_styles.cljs
```diff
@@ -524,6 +524,7 @@
     :warning [theme-colors/dark-tangerine-11 theme-colors/dark-tangerine-1]
     :error [theme-colors/red theme-colors/red-lightest]
     :info [theme-colors/blue theme-colors/blue-lightest]
+    :simple [theme-colors/white theme-colors/white]
```
jkorpelin, line 527: I think this is unnecessary as you shouldn't use the info-box when you don't want to see the borders?

jkorpelin (approved):
Other than the few small comments seems good to me

PR #623, opened 2021-06-15
TEET-1768 add mobile view 

tatut:


app/frontend/src/cljs/teet/asset/assets_view.cljs
```diff
@@ -441,18 +435,78 @@
...
+      [assets-results {:e! e! :atl atl :criteria criteria
+                       :asset-query query
+                       :results results
+                       :details (get-in app [:query :details])}]]]))
+
+(defn assets-page [e! {atl :asset-type-library :as app} state]
+  (if-not atl
+    [CircularProgress]
+    [context/provide :rotl (asset-type-library/rotl-map atl)
+     (if (responsivity-styles/mobile?)
```
jkorpelin, line 510: maybe instead of this we could use css breakponts to display none either of these? The onresize listener feels a bit inconsistent sometimes.

tatut: I don't want to render the heavy map/listing components twice, but it could perhaps be refactored

PR #643, opened 2021-06-17
TEET-1245/-1783 admin inspector tx info improvements

tatut:
- show tx time with millisecond precision
- show other tx metadata if no user associated with tx (any background task or integration tx)
- fix tx metadata for THK import
- add tx metadata for background notify

### DIFF
--- a/app/backend/src/clj/teet/notification/notification_ion.clj
+++ b/app/backend/src/clj/teet/notification/notification_ion.clj
@@ -41,7 +41,14 @@
          notifications-count (count tx-list)]
      (if (empty? tx-list)
        (log/info "No transaction info generated, automatic notifications skipped")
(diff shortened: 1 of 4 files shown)


kristjanZ:
What about migrations? Those don't also seem to have any "name" in the admin UI

tatut:
> What about migrations? Those don't also seem to have any "name" in the admin UI

Most normal migrations add schema items, so those can only come from migrations. And we know which one from schema entries.

But migration code that actually changes some data should add some tx metadata. But those are made in code so no simple generic way to enforce it.

aleksievelizhanin (approved):
Good that bucket and file name refactored from parameters to context. Code becomes more clear! Looks legit

PR #667, opened 2021-06-24
TEET-1714 - Ion for updating activity actual start/end dates

erno:


app/backend/src/clj/teet/thk/thk_integration_ion.clj
```diff
@@ -265,3 +266,61 @@
            csv->file
            dump))
   :ok)
+
+
+(def date-fixup-test-event
```
jkorpelin, line 271: Is this leftover from development?

erno: It's something to keep around when testing the code again later.

erno:
Closing for now as the immediate operational need has passed and a more generic "transact-anything" lambda could be implemented instead of this.

PR #670, opened 2021-06-24
TEET-1605 Create Cypress tests for Cost items workflow

outamaa:
This PR adds basic test suite for cost items CRUD workflow.


ci/browser-tests/cypress/integration/cost_items.spec.js
```diff
@@ -0,0 +1,210 @@
+describe("Cost items, totals and materials views", function() {
+  let costItemOid
```
tatut, line 2: should use context, not let variables that are set... as the cypress code runs it just creates a model of what to do in the test.

```cy.wrap(someExpression).as("myThing")```

that can later be used with `this.myThing`

PR #678, opened 2021-06-29
Teet 1771 contract authorization fn

jkorpelin:


app/backend/src/clj/teet/authorization/authorization_core.clj
```diff
@@ -0,0 +1,48 @@
...
+        seq
+        boolean)))
+
+(defn authorized-for-action?
+  "Check rights either based on the target or company
+  Given either company or a target find if the user has the right "
+  [{:keys [db user action target company]}]
+  {:pre [(or target company)
+         (not (and target company))]}
+  (let [authorized-roles (action @authorization-matrix)
```
tatut, line 31: should you assert that action is given and exists in the authorization matrix?
now giving `nil` as action will throw NPE

tatut: also it seems `db` is required, might as well pre check that is given

app/backend/src/clj/teet/authorization/authorization_core.clj
```diff
@@ -0,0 +1,48 @@
...
+                               set)
+        specific-roles
+        (cond target
+              (authorization-db/user-roles-for-target db user-ref target)
+              company
+              (authorization-db/user-roles-for-company db user-ref company))
+        roles (set/union user-global-roles specific-roles)]
+
+    (println "specific-roles: " specific-roles)
+    (println "authorized-roles " authorized-roles)
```
tatut, line 44: remove debug printlns

PR #685, opened 2021-07-01
TEET-1771 removes check for contract authorization because it cannot …

jkorpelin:
…work until you add people


aleksievelizhanin (approved):
Looks legit

PR #689, opened 2021-07-01
Teet 1651 authorization matrix

jkorpelin:
contract authorization support defrequest*


app/backend/src/clj/teet/user/user_db.clj
```diff
@@ -50,3 +51,30 @@
...
+     [?u :user/id _]
+     [?u :user/given-name ?given]
+     [(missing? $ ?u :user/deactivated?)]
+     [?u :user/family-name ?family]
+     [(str ?given " " ?family) ?full-name]
+     [(.toLowerCase ?full-name) ?full-name-lower]
+     [(.contains ^String ?full-name-lower ?search)]]])
+
+(defn find-user-by-name [db name]
+  (d/q '[:find (pull ?u [:db/id :user/id :user/given-name :user/family-name :user/email :user/person-id])
```
aleksievelizhanin, line 66: Different styles for (d/q ) function in one namespace are used. For example, user-info function put :user/person-id :db/id on new line, and this function put all keys in one line. The same style should be used across namespace (at least)

app/backend/src/clj/teet/user/user_db.clj
```diff
@@ -50,3 +51,30 @@
...
+         :where
+         (user-by-name ?u ?name)
+         :in $ % ?name]
+       db
+       user-query-rules
+       (str/lower-case name)))
+
+(defn find-all-users [db]
```
aleksievelizhanin, line 74: The same as the previous - all (d/q ..) functions should use the same style for keys - or put them all in one line, or, all should put some keys to new line as in (user-info) function

app/backend/src/clj/teet/user/user_db.clj
jkorpelin: They are on a different line there just because the line gets too long. I don't think we have followed any guidelines as the what attributes of the pull should be on different lines.

app/backend/src/clj/teet/db_api/core.clj
```diff
@@ -90,17 +91,43 @@
...
+                 (not (contains? options :project-id)))
+            (and unauthenticated?
+                 (not (contains? options :allowed-for-all-users?))
+                 (not (contains? options :authorization))
+                 (not (contains? options :contract-authorization))
+                 (not (contains? options :project-id)))
+            (and (not unauthenticated?)
+                 (or (contains? options :authorization)
+                     (contains? options :contract-authorization))
+                 (contains? options :project-id))
```
jkorpelin, line 112: This might be wrong?

PR #694, opened 2021-07-05
TEET-1857 hide edit button the user is not allowed to do the edit act…

jkorpelin:
…ion.

### DIFF
--- a/app/frontend/src/cljs/teet/contract/contract_partners_view.cljs
+++ b/app/frontend/src/cljs/teet/contract/contract_partners_view.cljs
@@ -17,13 +17,11 @@
             [teet.ui.form :as form]
             [teet.ui.select :as select]
             [teet.ui.common :as common]
-            [clojure.string :as str]
             [teet.ui.validation :as validation]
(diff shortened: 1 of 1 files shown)


app/frontend/src/cljs/teet/contract/contract_partners_view.cljs
```diff
@@ -498,17 +496,19 @@
...
+                               :params params
+                               :query {:page :edit-partner
+                                       :partner teet-id}})}
+       (tr [:buttons :edit])]]]))
 
 (defn partner-info
   [e! {:keys [params] :as app} selected-partner]
   [:div
-   [partner-info-header selected-partner params]
+   [partner-info-header selected-p]
```
aleksievelizhanin, line 511: "selected-p" is not very descriptive name. A suggest "selected-partner" instead

jkorpelin: nice catch I definitely didn't mean to change those params at all :D

aleksievelizhanin (approved):
Looks legit

PR #702, opened 2021-07-06
Browser tests snackbar success fix (branched from TEET-1802 PR, merge after that)

erno:


app/frontend/src/cljs/teet/contract/contracts_view.cljs
```diff
@@ -78,26 +83,27 @@
...
-                            :contracts-count (count contracts)
-                            :expand-contracts expand-contracts
-                            :collapse-contracts collapse-contracts}]
-     (doall
-       (for [contract (take show-count contracts)
-             :let [open? (@open-contracts (:db/id contract))]]
-         ^{:key (str (:db/id contract))}
-         [contract-card contract #(toggle-card (:db/id contract)) open?]))
-     [common-ui/scroll-sensor increase-show-count]]))
+  (when (not-empty contracts) ;; appease cypress test
```
jkorpelin, line 86: Can't see why the cypress test needs this. I think we would still want to show the contract header with the count of contracts even if there are no contracts found by the search.

app/frontend/src/cljs/teet/snackbar/snackbar_view.cljs
```diff
@@ -42,6 +42,11 @@
    [SnackbarContent
     {:style {:background-color (snack-color variant)
              :flex-grow 0}
+     :data-cy (str "snackbar-" (if (keyword? variant)
+                                 (name variant)
+                                 ;; else
+                                 (pr-str variant)
+                                 ))
```
jkorpelin, line 49: Should be cleaned up

PR #725, opened 2021-07-08
Playwright GHA config & trial test (for indexes)

erno:


tatut:
Please add a README file that documents how playwright is used.

As I understand, the recorded playwright tests might be more brittle than hand curated selectors in cypress tests.
It should be documented how to rerecord tests if they fail.

jkorpelin:
The playwright test failed, but the actions show that it succeeded?

jkorpelin:
And could we generate js instead of python for the tests? We already have js in the project and we would not need to have a new language in the project.

PR #726, opened 2021-07-08
Teet 1842 when authorized2

jkorpelin:


app/common/src/cljc/teet/authorization/authorization_check.cljc
```diff
@@ -92,6 +94,23 @@
...
+       (when-not (contains? contract-permissions cache-key)
+         (e! (common-controller/map->QueryUserAccess {:action action
+                                                      :entity entity
+                                                      :result-event (partial common-controller/->AuthorizationResult cache-key)}))))
+     (fn [user action entity contract-permissions component]
+       (let [cache-key [(:db/id user) action (:db/id entity)]]
+         (if (get contract-permissions cache-key)
+           component
+           (do
+             (log/debug "AUTHORIZED QUERY FAILED for action: " action)
```
tatut, line 111: won't this log this failure before the query has run?

app/frontend/src/cljs/teet/contract/contract_partners_view.cljs
```diff
@@ -468,6 +468,9 @@
        [Grid {:item true
               :xs 12
               :md 6}
+        [:p (pr-str "selected user: " selected-user)]
```
tatut, line 471: debugs here by design?

app/common/src/cljc/teet/authorization/authorization_check.cljc
jkorpelin: aa yea true, probably no need to log anything here, and just have a log at the authorizationresult.

PR #727, opened 2021-07-09
Set in format so that it works in AWS CLI v2

tatut:
Fix lambda invocation to work with AWS CLI v2 
see: https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration.html#cliv2-migration-binarypara


aleksievelizhanin (approved):
Looks legit

PR #731, opened 2021-07-09
Teet 1796

solita-erkkiv:


app/backend/src/clj/teet/task/task_commands.clj
```diff
@@ -95,6 +95,16 @@
             {:db/id (first pid)
              :file.part/status part-status})))
 
+(defn waiting-for-review-parts-tx
+  "Creates the tx records to update task part(s) from waiting for review to reviewing"
+  [task-parts]
+  (concat (map (fn [part]
```
tatut, line 101: extra `concat` here? what are we concatenating

app/common/src/cljc/teet/project/task_model.cljc
```diff
@@ -87,6 +87,14 @@
                 :else
                 :unassigned)))
 
+(defn any-task-part-waiting-for-review?
+  "Checks if any task parts under the task are in waiting for review status"
+  [task-parts]
+      (seq (filterv (fn [x]
```
tatut, line 93: instead of seq  and filter, you could use `some`

PR #752, opened 2021-07-28
TEET-1860 As any user I want to see related contracts (construction/owner's supervision) for a contract

outamaa:
The PR fixes the related contracts query.


aleksievelizhanin (approved):
Looks legic according to the AC that for construction activity contract the only related contract should be Safety Audit and Owner's supervision.

PR #765, opened 2021-07-30
TEET-1232 Adding new company employees to contract fixes

outamaa:
- Check if employee has already logged in when adding as a new user
- Check if employee has already been added to the contract


aleksievelizhanin (approved):
Looks legit to me

PR #767, opened 2021-07-30
TEET-1933 Fix "Add person view is closed when clicked on profile icon"

topipoh:
Do not pass the whole app state into add-personnel-form, so that the form is not re-rendered when :extra-panel-open? changes.


aleksievelizhanin (approved):
looks legit

PR #773, opened 2021-08-02
TEET-1874 As any user I want to see the partners representatives on the contract responsibilities page

outamaa:


aleksievelizhanin (approved):
looks legit

PR #782, opened 2021-08-05
Teet 1870 details view region

jkorpelin:
adds project region to details view of the project


aleksievelizhanin (approved):
looks legit to me

PR #786, opened 2021-08-05
Teet 1858 Contract key person documents

solita-erkkiv:


app/backend/src/clj/teet/file/file_commands.clj
```diff
@@ -85,6 +86,28 @@
     {:url (file-storage/upload-url key)
      :file (d/pull (:db-after res) '[*] file-id)}))
 
+(defcommand :file/upload-user-attachment
+  {:doc "Upload attachment file and optionally attach it to entity."
```
outamaa, line 90: Here we could document that we
- attach the file to user AND
- add it to the key person "application"

outamaa: This could even be named something like `upload-key-person-file` since this can't be used to just upload a file to the user.

app/backend/src/clj/teet/file/file_commands.clj
```diff
@@ -85,6 +86,28 @@
...
+  (let [key (new-file-key file)
+        res (tx [(merge (select-keys file file-keys)
+                          {:db/id "new-file"
+                           :file/s3-key key}
+                          (creation-meta user))])
+        file-id (get-in res [:tempids "new-file"])
+        _ (tx [{:db/id (contract-db/get-user-for-company-contract-employee db employee-id)
+                :user/files file-id}
+               {:db/id employee-id
+                :company-contract-employee/attached-files file-id}])]
```
outamaa, line 107: I'm not 100% sure, but I think you can combine the two transactions above by using the `"new-file"` placeholder also for `:user/files`.

app/frontend/src/cljs/teet/file/file_controller.cljs
```diff
@@ -197,10 +200,12 @@
...
                                  {:previous-version-id prev-file-id})
-                               (if attachment?
+                               (if user-attachment?
+                                 {:employee-id attach-to}
+                                (if attachment?
                                  {:project-id project-id
-                                  :attach-to attach-to}
-                                 {:task-id task-id}))
+                                    :attach-to attach-to}
+                                 {:task-id task-id})))
```
outamaa, line 208: Indentation seems a bit off above.

outamaa (changes requested):
A couple of small changes, otherwise looking good!

app/backend/src/clj/teet/file/file_commands.clj
solita-erkkiv: Renamed command and doc

app/backend/src/clj/teet/file/file_commands.clj
solita-erkkiv: Works! Thank you!

PR #788, opened 2021-08-06
TEET-1858 Fixes to document download and access rights

solita-erkkiv:


app/backend/resources/schema.edn
```diff
@@ -2222,4 +2222,14 @@
...
            :db/valueType :db.type/ref
            :db/cardinality :db.cardinality/one
-           :db/doc "Ref to company-contract-employee. Must be employee in this contract"}]]}]
+           :db/doc "Ref to company-contract-employee. Must be employee in this contract"}]]}
+
+ {:db/ident :migration/key-person-files
+  :txes [[{:db/ident :company-contract-employee/attached-files
+           :db/valueType :db.type/ref
+           :db/cardinality :db.cardinality/many
+           :db/doc "Reference to user/files that are linked to this key person"}
```
jkorpelin, line 2231: Isn't this just a list of files, and doesn't reference the user/files list? Since these are editable separately?

solita-erkkiv: I thought it would be better to use refs as you can then navigate in the admin inspect, no?

jkorpelin: Yes it's reference to the files but not the "user/files" specifically is it not?

PR #793, opened 2021-08-06
TEET-1956 Key person submitting for approval

solita-erkkiv:


app/frontend/src/cljs/teet/contract/contract_partners_view.cljs
```diff
@@ -495,15 +506,20 @@
...
          [(str/join ", " (mapv #(tr-enum %) (:company-contract-employee/role employee)))]
          [])
        [(if (true? key-person?)
           [:div {:style {:display :flex}}
-           [key-person-icon :gray]]
+           [key-person-icon (cond
+                              (du/enum= key-person-status :key-person.status/approval-requested) :yellow
+                              (du/enum= key-person-status :key-person.status/approved) :green
+                              (du/enum= key-person-status :key-person.status/rejected) :red
+                              :else :gray)]]
```
outamaa, line 522: I see the previous implementation handles color keywords, such as `:gray`, but instead of color, we could pass the status keyword to the `key-person-icon` component and act based on that.

outamaa: That is

```
[key-person-icon key-person-status]
```
and then instead of `case`ing by the color, we can `case` (or `cond`) based on the status itself.

outamaa (commented):
Looking good! Added a couple of refactoring suggestions.

